PR #10, opened 2024-12-19
Add UID and GID to set the user & group ID's during runtime

DB-Vincent:
This PR add the option to set UID/GID during runtime, making sure the application is running as a non-root user.
Besides this, it also helps with making sure UID/GID match between containers.


calibrain (approved):

calibrain:
Thank you for the work !
I was planning on adding support for non root, but you beat me to it, awesome !

PR #17, opened 2024-12-20
Download books from AA API if userkey is set

FlappyTaps:
This PR uses the Anna's Archive API to get a fast download URL to download the book directly. Bypasses parsing HTML with the Cloudflare proxy.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,7 @@ ENV CLOUDFLARE_PROXY_URL=http://localhost:8000
 ENV INGEST_DIR=/cwa-book-ingest
 ENV STATUS_TIMEOUT=3600
 ENV PYTHONPATH=/app
+ENV ANNAS_ARCHIVE_KEY=None
 
 # Default UID and GID (can be overridden at runtime)
 ENV UID=1000
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,6 @@ ENV CLOUDFLARE_PROXY_URL=http://localhost:8000
 ENV INGEST_DIR=/cwa-book-ingest
 ENV STATUS_TIMEOUT=3600
 ENV PYTHONPATH=/app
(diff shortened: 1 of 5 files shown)


Aymendje:
Thank you for the PR ! 
I was just looking into it and had an initial draft too :
https://github.com/calibrain/calibre-web-automated-book-downloader/pull/16/files
I don't mind dropping mine and merging yours, but could you change a bit the logic in book_manager.py to reuse basically what I have ?

This way, if for whatever reason we fail to get the book from the AA_API, we can still retry using the "old" method

calibrain:
Thank you for the PR !
I was just looking into it and had an initial draft too :
https://github.com/calibrain/calibre-web-automated-book-downloader/pull/16/files
I don't mind dropping mine and merging yours, but could you change a bit the logic in book_manager.py to reuse basically what I have ?

This way, if for whatever reason we fail to get the book from the AA_API, we can still retry using the "old" method

Dockerfile
```diff
@@ -17,6 +17,7 @@ ENV CLOUDFLARE_PROXY_URL=http://localhost:8000
 ENV INGEST_DIR=/cwa-book-ingest
 ENV STATUS_TIMEOUT=3600
 ENV PYTHONPATH=/app
+ENV AA_DONATOR_KEY=None
```
calibrain: Let's just leave this one unset

calibrain (changes requested):
Awesome ! One small change and then ill merge !

FlappyTaps:
Done :)

calibrain:
Oh, can you please completely remove it from the dockerfile ?

By default, if it exist (even if empty) it will be processed in python as `''` (empty string) and not as `None` so it will always try it, even if its empty

calibrain:
Awesome, thank you for the work !

PR #24, opened 2024-12-24
CF BYPASS

### LINKED ISSUE #18: Arm64 build?
Any chance of an ARM build? Running Docker on an Apple Silicon Mac here and ineant to avoid rosetta, etc.

calibrain:
The Calibre dependency was due to the script testing for validity of the downloaded file, as often they would be corrupted  from aa. But CWA is already doing that, so we are just having redundant code here.

For the cloudflarebypasser, I basically run my own version now, instead of depending on an external library, this way we have better control for debugging and on the docker image.

Fixes #18, #33, #27, #48, #65, #78, #86, #88, #89


Borriborri:
would it be a solution to use Flaresolverr? https://github.com/FlareSolverr/FlareSolverr
This is quite often used in other self hosted projects.

calibrain:
> would it be a solution to use Flaresolverr? https://github.com/FlareSolverr/FlareSolverr This is quite often used in other self hosted projects.

I had too many problems with Flaresolverr and other cloudflare bypasser, they are also really hard to debug, and not very flexible. By owning the code and running our own chrome instance, we can get logs of issues easier and better control on the code. Also it avoids useless network calls that have to be encoded and escaped to pass URL parameters between the project and the FlareSolverr instance

cloudflare_bypasser.py
```diff
@@ -0,0 +1,163 @@
...
+        logger.info("Bypass failed.")
+
+def _get_chromium_options(arguments: list[str]) -> ChromiumOptions:
+    options = ChromiumOptions()
+    for argument in arguments:
+        options.set_argument(argument)
+    return options
+
+def _genScraper() -> ChromiumPage:
+    arguments = [
```
mik593, line 114: Hi,

Following from  #18 I have been testing this a bit further on my Raspberry Pi. I had said it had trouble downloading, which appears to be the cloudless bypass falling (and it does then try moving on to alternatives) 

The failure seems to be with sandboxing - and chrome/chromium not starting up. Not that familiar but seems to be a a particular issue on containers.

The [current bypasser](https://github.com/mik593/CloudflareBypassForScraping/blob/7274bf650f2945a98fab4a5d37a87e1b0aa5e034/server.py#L69) adds the -no-sandbox flag if it’s being used within a docker container.

I added this on my own fork and it fixed this issue

Aymendje: You are correct !
Great catch, thank you !
I updated the PR to reflect that change
If I recall correctly, --no-sandbox is only needed if running as root, since I set a UID in my dev environment, I think thats why it worked, but you are 100% correct

Dockerfile
```diff
@@ -1,58 +1,47 @@
...
 HEALTHCHECK --interval=30s --timeout=30s --start-period=5s --retries=3 \
     CMD curl -f http://localhost:${FLASK_PORT}/request/api/status || exit 1
 
-# Entrypoint
-ENTRYPOINT ["/app/entrypoint.sh"]
-
-# Start application
-CMD ["python", "-m", "app"]
-
+ENTRYPOINT ["/usr/bin/dumb-init", "--"]
```
mik593, line 48: apologies if this is just this way for testing but thought I'd mention since I'd been running the branch - noticed these here don't start up the app right in the new version (no python -m app bit to trigger it.) but still works if changed to how it was before

Aymendje: Oh wow nice catch
Yeah I was running from a custom VSCode debug so forgot to properly test this after merging some upstream changes
Nice catch, thankyou !
I pushed the fix to the branch

calibrain: Oh wow nice catch
Yeah I was running from a custom VSCode debug so forgot to properly test this after merging some upstream changes
Nice catch, thankyou !
I pushed the fix to the branch

calibrain:
> is cloudflare bypass container needed when running the arm64 branch?

@Fuckingnameless  no, I dropped it and all the logic is now in the one container (actually, the cloudflare bypass was the main reason we were not arm compatible)

calibrain:
Should be something like this :

```
services:
    image: ghcr.io/calibrain/calibre-web-automated-book-downloader:CF_BYPASS
    environment:
      FLASK_PORT: 8084
      FLASK_DEBUG: true
      BOOK_LANGUAGE: en
    ports:
      - 8084:8084
    restart: unless-stopped
    volumes:
      - /tmp/data/calibre-web/ingest:/cwa-book-ingest
   ```

**Comment by bonny1992** [comment]
> Should be something like this :
> 
> ```
> services:
>   calibre-web-automated-book-downloader:
>     image: ghcr.io/calibrain/calibre-web-automated-book-downloader:CF_BYPASS
>     environment:
>       FLASK_PORT: 8084
>       FLASK_DEBUG: true
>       BOOK_LANGUAGE: en
>     ports:
>       - 8084:8084
>     restart: unless-stopped
>     volumes:
>       - /tmp/data/calibre-web/ingest:/cwa-book-ingest
> ```

I'm in the same boat as @Fuckingnameless 
```
calibre-web-automated-book-downloader  | 2025-02-15 02:48:45,006 - network - INFO - GET: https://annas-archive.org/slow_download/01bcc1ef0f1f52ea0c010af4e3cf94e3/0/2
calibre-web-automated-book-downloader  | 2025-02-15 02:48:45,116 - network - ERROR - Failed to fetch page: https://annas-archive.org/slow_download/01bcc1ef0f1f52ea0c010af4e3cf94e3/0/2, error: 403 Client Error: Forbidden for url: https://annas-archive.org/slow_download/01bcc1ef0f1f52ea0c010af4e3cf94e3/0/2
calibre-web-automated-book-downloader  | 2025-02-15 02:48:45,116 - network - INFO - GET: https://libgen.li/ads.php?md5=01bcc1ef0f1f52ea0c010af4e3cf94e3
```
Neither the `CF_BYPASS` tag of the docker image nor the `main` + cloudflare bypass container are working with "plain" Anna's urls.
I have to be lucky to be in the 5 first books from z-lib or if it's present on libgen.
I also tried putting the container behind vpns with no luck so far.

PR #32, opened 2025-01-04
Fix overflow for mobile, add spacing between buttons

cthorne:
Fix mobile being unable to scroll across to the buttons
Retain overflow-hidden when collapsed to hide content

Changing of CSS removes extra margin-bottom: rem padding incidentally


cthorne:
image

Issue #34, opened 2025-01-04
[Feature] Select multiple books to queue


llewy:
When searching for a specific author, I'd love to select multiple books to download all of them. Any chance this could be implemented?

Very nice piece of software b.t.w.!

yroyathon:
Just wanted to pipe in that I also agree this is a great app so far!   

For this particular feature request, I feel like I've seen the UI solution be to add a new column to the table containing a checkbox, and once one or more checkboxes are selected, a button near the top becomes active which when clicked does a specific action on those table rows that have a check.

PR #57, opened 2025-01-13
Select multiple books to queue

FunnyPocketBook:
Related to #34 

This PR adds:
- Additional column for checkboxes as the first column to select which books to download
- Select/deselect all books
- Button and modal to confirm download of selected books

Maybe the checkbox column should be at the end of the table, next to actions? 

image

image

image


calibrain:
Awesome work, thank you so much !

One small thing, when I merged #58, it created some merge conflicts with This PR, could you take a look ?
Sorry about that !

PR #64, opened 2025-01-14
Update readme.md env variable

FunnyPocketBook:
Adjust `AA_DONATOR_KEY` in text to reflect changes of the env variable


calibrain:
Oups goot catch

PR #66, opened 2025-01-22
fix: clear checkboxes on new search query

FunnyPocketBook:
This PR addresses an issue where the "download selected" functionality behaves unexpectedly. If a user makes a selection of books to download, performs a new search without clearing the previous selection, and selects new books, clicking "download selected" would still include the books from the prior selection. This causes the "download selected" button to fail, as the books from the previous selection are no longer available.

A potential enhancement could allow users to select books across multiple searches and download them collectively. This could be achieved by creating a dedicated section in the UI to track and display the currently selected books. This feature might go well with a broader UI overhaul, as has been suggested in some issues - maybe after your CF bypass rewrite?


calibrain:
Sounds good, thank you for the PR !

PR #69, opened 2025-01-25
feat: add commit date tagging to Docker image workflow

### LINKED ISSUE #68: Semantic versioning for the docker container
First I'd like to say that I'm really really happy with this project. Awesome work 🚀 

I can't be the only one that updates her repos automagically with renovate. Currently I'm using the 'latest' tag to pull the container, but that 'version' will of course never change and the only way to update is to manually restart the container. Having a semantic version for the container will kick renovate into action and create a pull request for me to deploy the new version (and yes, I am lazy, I automerge these PR's...)

It's as easy as pushing to a tag (see https://github.com/docker/metadata-action#semver for an example), but I'm sure it can also be scripted (I do that with my helm charts).

And if pushing to a tag really doesn't fit your workflow style, a simple date based tag like 20250125 would be awesome too!

cathelijne:
This PR adds a date-based tag to the docker container.

Example: 
Screenshot 2025-01-25 at 01 33 01

Fixes #68

### DIFF
--- a/.github/workflows/build-and-publish-docker-image.yml
+++ b/.github/workflows/build-and-publish-docker-image.yml
@@ -33,6 +33,7 @@ jobs:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           tags: |
             type=raw,value=latest,enable={{is_default_branch}}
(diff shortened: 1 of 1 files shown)


calibrain:
Awesome work, thank you !

PR #89, opened 2025-03-15
Enable proxy support via environment variables

lostwindsenril:
This update allows users to set the appropriate HTTP_PROXY and HTTPS_PROXY environment variables when starting the Docker container, and all outgoing HTTP/HTTPS requests will automatically use the specified proxy.

This change is particularly beneficial for users in restricted network environments (e.g., China), where access to certain online resources is often blocked.

### DIFF
--- a/network.py
+++ b/network.py
@@ -8,6 +8,9 @@ from typing import Optional
 from urllib.parse import urlparse
 from tqdm import tqdm
 
+import os
+proxy = os.environ.get("HTTP_PROXY") or os.environ.get("HTTPS_PROXY")
+
 from logger import setup_logger
 from config import MAX_RETRY, DEFAULT_SLEEP, CLOUDFLARE_PROXY, USE_CF_BYPASS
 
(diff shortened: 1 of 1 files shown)


network.py
```diff
@@ -80,11 +83,15 @@ def html_get_page_cf(url: str, retry: int = MAX_RETRY) -> str:
         return html_get_page(url, retry, skip_403=True)
     try:
         logger.info(f"GET_CF: {url}")
-        response = requests.get(
-            f"{CLOUDFLARE_PROXY}/html?url={url}&retries=3"
-        )
-        time.sleep(1)
```
aakashsethi20, line 86: Why remove the `time.sleep(1)?`

Though, not sure why it was there in the first place

calibrain:
Thank you for the PR ! It is a great idea, I have added it to  #24  and included proper CONFIG and Readme updates !

network.py
calibrain: The sleep is to not trigger the anti bot detection of cloudflare by spamming it too much too quickly

lostwindsenril: It was just my oversight, thanks for the reminder!

PR #143, opened 2025-04-18
Fix cwa-book-downloader Permission Issues

l1berate:
Fix permission issues that occur if user forgets to create the `cwa-book-downloader` folder before creating the docker container since docker makes the directory with root ownership.


calibrain:
Another good catch ! Thank you !

Issue #120, opened 2025-04-06
Clowdflare bypass always failing


bevtrin:
Hi,

I unfortunately can't get this working. the Cloudflare bypass continuously fails. I have it running behind gluetun. 

Log:
```
2025-04-06 08:51:50,802 - cloudflare_bypasser - INFO - cloudflare_bypasser.py:101 - Starting Cloudflare bypass... Retry: 1 / 3
2025-04-06 08:51:50,803 - cloudflare_bypasser - INFO - cloudflare_bypasser.py:106 - Attempt 1: Verification page detected. Trying to bypass...
2025-04-06 08:52:03,603 - cloudflare_bypasser - DEBUG - cloudflare_bypasser.py:66 - Verification button found. Attempting to click.
2025-04-06 08:52:10,469 - cloudflare_bypasser - INFO - cloudflare_bypasser.py:101 - Starting Cloudflare bypass... Retry: 2 / 3
2025-04-06 08:52:10,469 - cloudflare_bypasser - INFO - cloudflare_bypasser.py:106 - Attempt 2: Verification page detected. Trying to bypass...
2025-04-06 08:52:10,529 - werkzeug - INFO - _internal.py:97 - 127.0.0.1 - - [06/Apr/2025 08:52:10] "GET /request/api/status HTTP/1.1" 200 -
2025-04-06 08:52:15,644 - cloudflare_bypasser - DEBUG - cloudflare_bypasser.py:66 - Verification button found. Attempting to click.
2025-04-06 08:52:22,221 - cloudflare_bypasser - INFO - cloudflare_bypasser.py:101 - Starting Cloudflare bypass... Retry: 3 / 3
2025-04-06 08:52:22,221 - cloudflare_bypasser - INFO - cloudflare_bypasser.py:106 - Attempt 3: Verification page detected. Trying to bypass...
2025-04-06 08:52:26,757 - werkzeug - INFO - _internal.py:97 - 172.23.0.1 - - [06/Apr/2025 08:52:26] "GET /request/api/status HTTP/1.1" 200 -
2025-04-06 08:52:34,941 - cloudflare_bypasser - DEBUG - cloudflare_bypasser.py:66 - Verification button found. Attempting to click.
2025-04-06 08:52:40,695 - werkzeug - INFO - _internal.py:97 - 127.0.0.1 - - [06/Apr/2025 08:52:40] "GET /request/api/status HTTP/1.1" 200 -
[...]

PR #146, opened 2025-04-19
Stop Gunicorn from Giving Up on Slow Tor Requests

l1berate:
I tried running the docker-compose.yml referenced in this comment https://github.com/calibrain/calibre-web-automated-book-downloader/issues/120#issuecomment-2816539268 from issue #120 and was able to reproduce the same 'Server is not responding' error. I found the issue would happen whenever the captcha bypass took more than 30 seconds due to tor's slower connection. Adding -t 300 would let gunicorn wait up to 5 minutes now.


calibrain:
Awesome find ! Thank you !

PR #161, opened 2025-04-23
use PUID and PGID env vars

RyantHults:


holzerseb:
Would love if this PR gets resolved and approved. I just found out that I've been running my containers as root because I was using the default PUID/GUID env vars and CWABD uses UID and GID. I've now changed my docker-compose to just define both PUID and UID so it should work in both cases.

alexhb1:
Happy to revisit this, I'll see if we can get it added without breaking existing installs :)

alexhb1:
I've took a stab at implementing this in a future PR. Expected behavior will be as follows: 

  - Existing installs with no values set: Use 1000:1000
  - Existing installs with UID/GID set: Use existing values
  - New installs: Use PUID/PGID (documented as recommended)
  - Log messages: Gently remind current users to migrate, but no issues if not

PR #169, opened 2025-04-28
Do not fail on folder permission error

calibrain:


opswhisperer:
This issue is weird because my setup has not changed but this this comes and goes for me. The current image crashes with permission errors but the previous one did not.  Also

```
Folder /cwa-book-ingest is not writable, changing ownership
Changing ownership of /cwa-book-ingest to root:100
chown: changing ownership of '/cwa-book-ingest': Operation not permitted
Failed to change user ownership for /cwa-book-ingest, continuing...
chown: changing group of '/cwa-book-ingest': Operation not permitted
Failed to change group ownership for /cwa-book-ingest, continuing...
chmod: changing permissions of '/cwa-book-ingest': Operation not permitted
```

What is the test for writable? I can exec into this pod and the user (root in this case) can write to the directory just fine.

opswhisperer:
new image works either way `Failed to change group permissions for /cwa-book-ingest, continuing...`

opswhisperer:
```
root@dlbooks-7bb76bdd8d-4cl7k:/app# echo 0123456789_TEST | sudo -E -u "appuser" HOME=/app tee /cwa-book-ingest/calibre-web-automated-book-downloader_TEST_WRITE > /dev/null
sudo: unknown user appuser
sudo: error initializing audit plugin sudoers_audit
```

looking at `/etc/shadow/` this container does not have a user `appuser`

calibrain:
oh, I found the issue, it was unrelated
Fixed in #170 
I just saw you were running as root anyway so this should have never failed for you in the first case

Issue #56, opened 2025-01-13
[FEATURE] Add authentication 


tobiraak:
Love your work! I would love to make the downloader available via my server that is accessible through the internet so it would be great to have some kind of authentication when exposing the downloader. Would this be possible and an overall interest?

Aymendje:
To be honest, I dont want to mess with users, its a a whole can of worms
For my personal case, I use cloudflare + authentik as my SSO that forces authentication first. It works well, and you could also use a reverse proxy to setup auth

I will leave this feature request open, if anyone wants to take a stab at it, or if there is more discussion around this, but I dont have plans on fully supporting users yet.
Maybe I could just reuse whatever database calibre-web has and read users from there ?
its something easier and doable

calibrain:
To be honest, I dont want to mess with users, its a a whole can of worms
For my personal case, I use cloudflare + authentik as my SSO that forces authentication first. It works well, and you could also use a reverse proxy to setup auth

I will leave this feature request open, if anyone wants to take a stab at it, or if there is more discussion around this, but I dont have plans on fully supporting users yet.
Maybe I could just reuse whatever database calibre-web has and read users from there ?
its something easier and doable

Bat-Shark:
I would also want this to have a form of auth before exposing it to users or even myself remotely. I have several services behind a reverse proxy with caddy already but still require auth. I'm would also be perfectly fine if it reuses the users and auth from calibre-web since that has everything already.

gowtham1337:
> To be honest, I dont want to mess with users, its a a whole can of worms For my personal case, I use cloudflare + authentik as my SSO that forces authentication first. It works well, and you could also use a reverse proxy to setup auth
> 
This is a perfect solution!

tobiraak:
Okay I implemented a basic nginx authentication with a .htpasswd file and username/password.
Should be good enough I guess, making use of the CWA users would of course be incredible. If your library would be integrated into CWA that would be nice, direct search and download inside the application.
If you want you can close the issue now or leave it open for further features.

calibrain:
Awesome
I will leave it open for now as end goal to be able to connect through whatever CWA is using already !

tobiraak:
Hey, just wanted to know if you have any updates on this or any features planned?

calibrain:
As I maintain this project on my free time, I don't have a road map or anything yet
I might add in the future some access to the CWA database, but its probably not short term
Maybe around end of this year if I get some free time ?

PR #181, opened 2025-05-29
Add simple authentication

trallen:
This commit adds user authentication, using Calibre-Web's app.db as its authentication source, as requested in #56. It uses @prinzpi's [comment](https://github.com/calibrain/calibre-web-automated-book-downloader/issues/56#issuecomment-2919335169) as a starting point, but integrates the logic directly into the app.

This requires the user to specify the environment variable CONFIG_ROOT, set to Calibre-Web's config directory (the directory that contains the app.db database that holds the user's authentication information).

If the user does not wish to add authentication, they can simply leave CONFIG_ROOT unset, or not pointing at Calibre-Web's app.db directory.


env.py
```diff
@@ -4,6 +4,8 @@
 def string_to_bool(s: str) -> bool:
     return s.lower() in ["true", "yes", "1", "y"]
 
+CONFIG_ROOT = Path(os.getenv("CONFIG_ROOT", "/config"))
```
calibrain: Lets remove CONFIG_ROOT and only keep DB_PATH for now

app.py
```diff
@@ -287,6 +300,58 @@ def internal_error(error: Exception) -> Union[Response, Tuple[Response, int]]:
...
+    Authentication middleware that:
+    1. Validates Basic Auth credentials against SQLite database
+    2. Forwards authenticated requests to backend service
+    3. Adds X-Authenticated-User header for backend to identify the user
+
+    Database requirements:
+    - Table 'user' with columns: 'name' (username), 'password'
+    - Passwords must be hashed with Werkzeug's generate_password_hash()
+    """
+    if not os.path.isdir(CONFIG_ROOT) or not os.path.isfile(DB_PATH):
```
calibrain: Lets remove CONFIG_ROOT and only keep DB_PATH

app.py
```diff
@@ -73,6 +82,10 @@ def index() -> str:
     """
     Render main page with search and status table.
     """
+    if os.path.isdir(CONFIG_ROOT) or os.path.isfile(DB_PATH):
```
calibrain: Lets remove CONFIG_ROOT and keep only DB_PATH

app.py
```diff
@@ -2,15 +2,17 @@
...
-from werkzeug.wrappers import Response 
+from werkzeug.security import check_password_hash
+from werkzeug.wrappers import Request, Response
 from flask import url_for as flask_url_for
 import typing
 
 from logger import setup_logger
 from config import _SUPPORTED_BOOK_LANGUAGE, BOOK_LANGUAGE
-from env import FLASK_HOST, FLASK_PORT, APP_ENV, DEBUG
+from env import FLASK_HOST, FLASK_PORT, APP_ENV, CONFIG_ROOT, DB_PATH, DEBUG
```
calibrain: Lets remove CONFIG_ROOT and keep only DB_PATH

env.py
```diff
@@ -4,6 +4,8 @@
...
+DB_PATH = Path(os.getenv("DB_PATH", CONFIG_ROOT / "app.db"))
```
calibrain: Could we 
1. Rename DB_PATH to CWA_DB_PATH?
2. If not set, lets make it None instead

readme.md
```diff
@@ -60,9 +60,12 @@ An intuitive web interface for searching and requesting book downloads, designed
 | `TZ`              | Container timezone      | `UTC`              |
 | `UID`             | Runtime user ID         | `1000`             |
 | `GID`             | Runtime group ID        | `100`              |
+| `CONFIG_ROOT`     | Calibre-Web's /config   | None               |
```
calibrain: Replace by CWA_DB_PATH

readme.md
```diff
@@ -60,9 +60,12 @@ An intuitive web interface for searching and requesting book downloads, designed
...
 | `ENABLE_LOGGING`  | Enable log file         | `true`             |
 | `LOG_LEVEL`       | Log level to use        | `info`             |
 
+If you wish to enable authentication, you must set CONFIG_ROOT to point to the directory containing Calibre-Web's app.db, in order to match the username and password.
```
calibrain: DB_PATH

readme.md
```diff
@@ -165,6 +168,7 @@ volumes:
 ```yaml
 volumes:
   - /your/local/path:/cwa-book-ingest
+  - /cwa/config/path:/config
```
calibrain: DB_PATH

templates/auth.html
calibrain, line 1: Do you think you could make it into a BASIC auth instead of adding an auth.html file ?
I want to avoid maintaining extra UI elements for now

calibrain (changes requested):
Hi ! Thank you so much for this ! 
Sorry for the delay in the review, I was away traveling, and just came back
I have a few comments if you have some time for an update :

Instead of having to env variable for essentially the same thing, could we either
a) only use a CWA_BD_PATH, and check for the DB file there instead.
b) Have a ahrdcoded path we check for, (ie `/cwa_bd.db`) and on initialization, check if that path exists and if try to load the db, if it fails, ignore auth, it if succeds start checking for auth ?

Also, in general, I want to avoid adding front end elements that needs to be maintained. Could we instead do a BASIC AUTH ? 

In the future I plan a complete rewrite of the front end, but until then, I want to limit tech debt

readme.md
calibrain: Also, lets make sure to add a `:ro` flag so its read only
`- /cwa/config/app.db:/cwa-db.db:ro`

templates/auth.html
trallen: Sure, I can understand that! Do you want a native implementation, or should I use [Flask-HTTPAuth](https://github.com/miguelgrinberg/Flask-HTTPAuth) (or some other library)?

trallen: For the moment, I've written a simple native implementation: a 401 response with a [WWW-Authenticate](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/WWW-Authenticate) header. I think this will be sufficient.

calibrain:
Thank you so much for the hard work !

PR #187, opened 2025-06-16
Bypass with flaresolver

FedeDR:
Remove the internal Cloudflare bypass to use an external one


balgerion:
any chance to merge this pr ? : )

jaannnis (approved):

Hustleberry:
And chance of putting up a new build with thise fixes?

calibrain:
Hi, sorry I wasn't available for a while
I want to keep cwabd as standalone as possible, and had issues with flaresolvarr and other in the past, and that is one of the reasons why we went with a single docker with more control on the binaries and retry logic.

If you really want / prefer flaresolverrr and you already have a setup using it with other projects, I would be open to (yet another) env flag that if set would use flaresolverrr instead. You can create another Dockerfile that would not build chromium (a bit like we have for the tor version)

Hustleberry:
The solution to get at least the possibility to utlize an existing flarsolver instance would be quite nice honestly - I appreciate this offer and for sure I can understand why you want to keep your container standalone.

If maken utilization of an already existing flaresolverr instance would not be too much of a hassle, I would be happy with that solution :)

sirjmann92:
Any update on this? The cloudflarebypass image is massive compared to flaresolverr, plus it's a waste of space entirely if we have a working solution. I would prefer a modular approach, but I'll take a working solution over a non-functioning one any day.

chrisfu:
> If you really want / prefer flaresolverrr and you already have a setup using it with other projects, I would be open to (yet another) env flag that if set would use flaresolverrr instead. You can create another Dockerfile that would not build chromium (a bit like we have for the tor version)

I'd back this solution. Default to internal and optional use of Flaresolverr will absolutely keep the devil from the door. Given that right now, the internal solution is failing, I think this would be the optimal and most flexible solution until the internal solver is fixed.

cthorne:
I've tested this solution on a separate fork (via Docker build actions), I do like the optional env overwrites to offload onto another service - less maintenance in this repo.

Mine utilised Byparr as according to comments on Flaresolverr it may be more up to date / useful (can't personally comment, haven't reviewed eithers code)

yroyathon:
I would definitely/immediately use an env flag that allowed me to point to Byparr since I'm already running that.  I've stopped using Flaresolverr since that stopped being as effective.

pabsi:
Was pointed here from https://github.com/calibrain/calibre-web-automated-book-downloader/issues/182#issuecomment-3215738630

I have to say I'd personally take a different approach on this. Like @chrisfu said: we should keep both options: default built-in cloudflare_bypass, and optionally use an external one (flaresolverr, byparr, etc) via an env var.

This PR completely removes cloudflare_bypass. Maybe the default approach is good for those who do not have an external cf bypass mechanism. Hence I'd have both (if possible).

FedeDR:
I'm closing this PR in favor of https://github.com/calibrain/calibre-web-automated-book-downloader/pull/245

PR #189, opened 2025-06-26
Fix: Correct healthcheck behavior

jojo141185:
### Problem
The current `HEALTHCHECK` uses a custom script that fails on any non-2xx HTTP response. When the API returns a `401 Unauthorized` status (a common and valid case), the container is incorrectly marked as `unhealthy`, even though the web server is running perfectly.

### Solution
This PR replaces the custom `pyrequests` script with a standard `curl` command. The new `HEALTHCHECK` now only verifies that the server is reachable and responding to connections, regardless of the HTTP status code.

`curl` is used instead of patching the existing script because it is the standard, most robust, and simplest solution for this type of check in a Docker environment. It removes a custom dependency and follows a well-understood pattern.

This change ensures the container's health status accurately reflects the state of the server process, improving reliability for orchestration platforms.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,6 +34,8 @@ RUN apt-get update && \
     apt-get install -y --no-install-recommends \
     # For locale
     locales tzdata \
+    # For healthcheck
+    curl \
     # For entrypoint
     dumb-init \
     # For dumb display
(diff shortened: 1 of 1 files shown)


mruijzendaal:
@calibrain This looks like a beautiful fix. Could you please merge this?

In the meantime, I've edited my `docker-compose.yml` to build from this PR instead of using the pre-compiled docker image:
```diff
services:
  calibre-web-automated-book-downloader:
-    image: ghcr.io/calibrain/calibre-web-automated-book-downloader:latest
+    build:
+      context: https://github.com/jojo141185/calibre-web-automated-book-downloader.git#fix-healthcheck
```

edit: specify branch in the build context.

PR #205, opened 2025-07-30
use the FLASK_HOST variable as declared in the documentation

ccarpo:
I noticed that the FLASK_HOST was not used as stated in the documentation. I need it so that FLASK is listening on my IPv6 address.


calibrain:
Good catch !

PR #215, opened 2025-08-14
Add semantic versioning for the docker images

nuovamaila:
For every push of a tag, create an image with the same tag.

### DIFF
--- a/.github/workflows/build-and-publish-docker-image.yml
+++ b/.github/workflows/build-and-publish-docker-image.yml
@@ -6,6 +6,9 @@ on:
       - 'main'
     tags:
       - 'v*'
+  pull_request:
+    branches:
+      - 'main'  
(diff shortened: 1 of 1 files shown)


calibrain:
@nuovamaila thank you for this PR !
Just wondering, how does it work for a normal workflow ?
Do we need to manually deploy to specific branches ?
Cant we just have a SEMVER file somewhere and auto increment it and auto tag releases ?

nuovamaila:
@calibrain Hi, the workflow activates every time a new tag is created for the repository. So, you can create a new release/tag (such as `v1.0.0`) and then an image with the corresponding tag will be created

calibrain:
Sounds good, thank you ! I will merge and release a version v0.1.0 and can you verify if it works for you !

PR #231, opened 2025-08-19
Add Download Queue Management & Concurrent Downloads

stephondoestech:
## Summary

  This PR completely overhauls the download system to support concurrent downloads, cancellation, and advanced queue
  management. No more waiting for stuck downloads to block the entire queue!

  ## New Features

  ###  **Concurrent Downloads**
  - **3 simultaneous downloads** by default (configurable)
  - ThreadPoolExecutor-based architecture replaces single-threaded bottleneck
  - Downloads no longer block each other

  ###  **Download Cancellation**
  - **Cancel button** for active/queued downloads
  - Clean cancellation with partial file cleanup
  - Thread-safe cancellation flags

  ### **Queue Priority & Reordering**
  - **Priority-based queue** (lower number = higher priority)
  - **Editable priority inputs** in UI for queued items
  - **Bulk reordering** API support

  ## UI Enhancements

  - **Enhanced status table** with Priority, Progress, and Actions columns
  - **Progress bars** for active downloads
  - **Cancel buttons** for downloads
  - **Queue management controls**: Refresh, Clear Completed, Active Counter
  - **Real-time updates** every 60 seconds

### New Architecture
  - `queue.PriorityQueue` replaces simple `set()` for proper ordering
  - `concurrent.futures.ThreadPoolExecutor` handles worker threads
  - `threading.Event` flags enable clean cancellation
  - Thread-safe operations with proper locking

  ### New API Endpoints
  DELETE /api/download/{id}/cancel     # Cancel downloads
  PUT /api/queue/{id}/priority        # Set priorityPOST /api/queue/reorder             # Bulk reorder
  GET /api/downloads/active          # Active tracking
  DELETE /api/queue/clear            # Clear completed

  ### New Environment Variables
  ```bash
  MAX_CONCURRENT_DOWNLOADS=3          # Worker thread count
  DOWNLOAD_PROGRESS_UPDATE_INTERVAL=5  # Progress frequency
  ```

### Fixes

  - Resolves: Downloads blocking when one gets stuck
  - Resolves: No way to cancel problematic downloads
  - Resolves: No queue management or prioritization
  
  #183

PR #243, opened 2025-08-24
Rework UI to add inline status

RyantHults:
# Summary

This PR reworks the UI to merge the search and downloads sections. 
- Each row now has a "status" column. This column is empty until an entry is downloaded. the download button is disabled until the file has finished downloading.
- removed separate download section
- removed Accordion div for search section

## Examples
Screenshot from 2025-08-23 15-37-07
Screenshot from 2025-08-23 15-27-49

## Notes
- I did not include the book ID in the status column, as I don't really know how useful that is to the average user and it would make the column quite large.
- I'm not sure if there is a reason the download button should be re-enabled if the book successfully downloaded. I left it as is for now, but wanted to get some thoughts on this.

### DIFF
--- a/static/css/styles.css
+++ b/static/css/styles.css
@@ -533,12 +533,7 @@ h1, h2, h3, h4, h5, h6,
     color: var(--heading-color) !important;
 }
 
-/* Ensure accordion titles are visible */
-.uk-title {
-    color: var(--heading-color) !important;
-}
-
 /* Ensure search results heading is visible */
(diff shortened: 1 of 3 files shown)


calibrain:
Hey ! Thank you very much for this PR ! 
But I think it might already be (partially)solved  by #231

calibrain:
Closing for now, feel free to update conflicts and submit a new one !
Thank you

PR #244, opened 2025-08-25
Display download progress in the ui

davidemarcoli:
This pr adds the download progress to the book model and displays it in the download status section in the ui.


calibrain:
Hi ! Thank you so much for your contribution ! 
But I think it might already be already solved  by #231

davidemarcoli:
@calibrain Oh, you're right! I closed the pr!

PR #266, opened 2025-09-02
Display book covers at full height

RyantHults:
# Why
Book covers in the UI are currently cut off on the top and bottom, making it hard to see.

# How
doubled the height of the book cover image div so the covers are not cut off. I found that setting it to a specific size (rather than `h-full`) resulted in better handling of small images and made for a more consistent look.

# Before
Screenshot from 2025-09-02 11-44-49
Screenshot from 2025-09-02 11-45-06

# After
Screenshot from 2025-09-02 11-41-41
Screenshot from 2025-09-02 11-44-38

### DIFF
--- a/static/js/main.js
+++ b/static/js/main.js
@@ -94,8 +94,8 @@
 
   // ---- Cards ----
   function renderCard(book) {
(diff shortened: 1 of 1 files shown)


calibrain:
Awesome thank you :D !

Issue #122, opened 2025-04-12
[FEATURE] Separate download folders


binarygeek119:
I have calibre web automated set up 3 time one for each type of media I have books/audiobooks, comics, and magazines 

If we could tell the app this media gets downloaded to this ingest folder

If I download a magazine then I can set up it to download in Ingest-Magazines 

If I download a comic book then I can set up it to download in Ingest-comics

I see this being as simple as category type = folder 

Not by file extension 

Audiobook=/media/books
Book=/media/books
Comic=/media/comics
Magazine=/media/magazines
Paper=/media/papers
Other=/media/other

TigreModerata:
Hi, I'd like to try to contribute (first timer), if that's ok? 
I'm not sure how you'd like this to go: a list of options above/next to the download button on the thumbnail? Or maybe show the options only if Details is opened, while that the thumbnail download always goes to a default-folder? 
*edit: I saw you can search by type - Anna's archive does have the content type info, but it's not a property of BookInfo I think. I'm guessing that's the task: automate recovery of the content type and sorting into appropriate folders where possible.

PR #297, opened 2025-11-04
[FEATURE] Separate download folders #122

TigreModerata:
Re: Issue #122
Fetches content type from search results - displays it on thumbnails in results grid;
Fetches content type from book id detail page (dfaults to "Other") and uses it to construct the `final_path`.

### DIFF
--- a/logger.py
+++ b/logger.py
@@ -13,21 +13,25 @@ class CustomLogger(logging.Logger):
     def error_trace(self, msg: Any, *args: Any, **kwargs: Any) -> None:
         """Log an error message with full stack trace."""
         self.log_resource_usage()
+        kwargs.pop('exc_info', None)
         self.error(msg, *args, exc_info=True, **kwargs)
 
     def warning_trace(self, msg: Any, *args: Any, **kwargs: Any) -> None:
         """Log a warning message with full stack trace."""
         self.log_resource_usage()
+        kwargs.pop('exc_info', None)
         self.warning(msg, *args, exc_info=True, **kwargs)
     
     def info_trace(self, msg: Any, *args: Any, **kwargs: Any) -> None:
         """Log an info message with full stack trace."""
         self.log_resource_usage()
+        kwargs.pop('exc_info', None)
         self.info(msg, *args, exc_info=True, **kwargs)
     
     def debug_trace(self, msg: Any, *args: Any, **kwargs: Any) -> None:
         """Log a debug message with full stack trace."""
         self.log_resource_usage()
+        kwargs.pop('exc_info', None)
         self.debug(msg, *args, exc_info=True, **kwargs)
     
     def log_resource_usage(self):


calibrain:
Hi, thank you very much for the PR
I will review it during the weekend, and will approve then !

calibrain:
Hi, sorry it took so long to review, I had a few busy days

This looks good, but one thing
Can we make it that it's an opt-in ?

Ideally I want to avoid changing the current way it works for people (some people are using CWA-BD outside of CWA and are not expecting new folders)

instead, can we make each "type" opt in with an env variable ?
For example, if I set something like `INGEST_DIR_COMIC`, then if its a comic, use the new dir, if not, use the default one (`INGEST_DIR`) ?

Thank you !

TigreModerata:
Hi, thanks for the feedback! Makes sense. I'll get to it this week, probably the weekend.

TigreModerata:
@calibrain Updated the PR, changed it to use env variables: if set, `INGEST_DIR_x` will be the name of the subfolder to store content of type x in; if no name is set for a specific type, the content will go into the default folder that `INGEST_DIR` is bound to. 
I've added this explanation to the readme.

backend.py
```diff
@@ -135,6 +123,16 @@ def _book_info_to_dict(book: BookInfo) -> Dict[str, Any]:
         if value is not None
     }
 
+def _prepare_download_folder(book_info: BookInfo) -> Path:
+    """Prepare final content-type subdir"""
+    content = book_info.content
+    content_type = [x for x in ContentType if x.value in content]
+    dir_name = eval("INGEST_DIR_" + content_type[0].name) if content_type else None
```
calibrain: Please avoid using `eval` whenever possible, instead you can use a dictionary or something

env.py
```diff
@@ -4,19 +4,20 @@
...
 TMP_DIR = Path(os.getenv("TMP_DIR", "/tmp/cwa-book-downloader"))
 INGEST_DIR = Path(os.getenv("INGEST_DIR", "/cwa-book-ingest"))
+INGEST_DIR_BOOK_FICTION = Path(os.getenv("INGEST_DIR_BOOK_FICTION", ""))
+INGEST_DIR_BOOK_NON_FICTION = Path(os.getenv("INGEST_DIR_BOOK_NON_FICTION", ""))
+INGEST_DIR_BOOK_UNKNOWN = Path(os.getenv("INGEST_DIR_BOOK_UNKNOWN", ""))
+INGEST_DIR_MAGAZINE = Path(os.getenv("INGEST_DIR_MAGAZINE", ""))
+INGEST_DIR_COMIC_BOOK = Path(os.getenv("INGEST_DIR_COMIC_BOOK", ""))
+INGEST_DIR_AUDIOBOOK = Path(os.getenv("INGEST_DIR_AUDIOBOOK", ""))
+INGEST_DIR_STANDARDS_DOCUMENT = Path(os.getenv("INGEST_DIR_STANDARDS_DOCUMENT", ""))
+INGEST_DIR_MUSICAL_SCORE = Path(os.getenv("INGEST_DIR_MUSICAL_SCORE", ""))
```
calibrain: When you use Path on an empty string, it returns "." which is a valid current directoy
Ideally we want to avoid that, and default to INGEST_DIR instead. You could do it like this for example :
(also this way, you only need to import `DOWNLOAD_PATHS`. You could also just write a small function in env.py to run a switch case

```suggestion
INGEST_DIR_BOOK_FICTION = os.getenv("INGEST_DIR_BOOK_FICTION", "")
INGEST_DIR_BOOK_NON_FICTION = os.getenv("INGEST_DIR_BOOK_NON_FICTION", "")
INGEST_DIR_BOOK_UNKNOWN = os.getenv("INGEST_DIR_BOOK_UNKNOWN", "")
INGEST_DIR_MAGAZINE = os.getenv("INGEST_DIR_MAGAZINE", "")
INGEST_DIR_COMIC_BOOK = os.getenv("INGEST_DIR_COMIC_BOOK", "")
INGEST_DIR_AUDIOBOOK = os.getenv("INGEST_DIR_AUDIOBOOK", "")
INGEST_DIR_STANDARDS_DOCUMENT = os.getenv("INGEST_DIR_STANDARDS_DOCUMENT", "")
INGEST_DIR_MUSICAL_SCORE = os.getenv("INGEST_DIR_MUSICAL_SCORE", "")

DOWNLOAD_PATHS = {
    "BOOK_FICTION": Path(INGEST_DIR_BOOK_FICTION) if INGEST_DIR_BOOK_FICTION else INGEST_DIR,
    "BOOK_NON_FICTION": Path(INGEST_DIR_BOOK_NON_FICTION) if INGEST_DIR_BOOK_NON_FICTION else INGEST_DIR,
    "BOOK_UNKNOWN": Path(INGEST_DIR_BOOK_UNKNOWN) if INGEST_DIR_BOOK_UNKNOWN else INGEST_DIR,
    "MAGAZINE": Path(INGEST_DIR_MAGAZINE) if INGEST_DIR_MAGAZINE else INGEST_DIR,
    "COMIC_BOOK": Path(INGEST_DIR_COMIC_BOOK) if INGEST_DIR_COMIC_BOOK else INGEST_DIR,
    "AUDIOBOOK": Path(INGEST_DIR_AUDIOBOOK) if INGEST_DIR_AUDIOBOOK else INGEST_DIR,
    "STANDARDS_DOCUMENT": Path(INGEST_DIR_STANDARDS_DOCUMENT) if INGEST_DIR_STANDARDS_DOCUMENT else INGEST_DIR,
    "MUSICAL_SCORE": Path(INGEST_DIR_MUSICAL_SCORE) if INGEST_DIR_MUSICAL_SCORE else INGEST_DIR,
    "DEFAULT" : INGEST_DIR,
}

```

calibrain:
Hi ! Awesome, thank you !
Left 2 smaller comment, and I think you might need to resolve some merge conflicts, but looks great
Thank you again !

env.py
TigreModerata: The reason the default was not INGEST_DIR was that in the code the final dir becomes `IGEST_DIR / INGEST_DIR_COMIC_BOOK`. If not set we'd have `INGEST_DIR/`   
That way the user can just set the names of the various folders, rather than having to set the complete path (with parent folder), as not all users will necessarily know the correct docker volume to use... 
I did not know that "" goes to . in Path though, so yeah, it has to be changed... Do you mind if I use your suggestion, with download_paths, but use an empty string for the else?
`"MUSICAL_SCORE": Path(INGEST_DIR_MUSICAL_SCORE) if INGEST_DIR_MUSICAL_SCORE else "",`

calibrain: I think if we are giving this much flexibility to users, might as well give them the full path flexibility ?
For example, if someone is running 2-3 different versions of CWA, and running a single CWA-BD, they might want to download to fully different paths ?
(ie : /data/cwa-book/ingest, and then /data/cwa-comic/ingest) etc. ?

What do you think ?

TigreModerata: I think I may be missing something :D It looked to me like INGEST_DIR is the name of the dir on the docker container that is bound to /tmp/data/calibre-web/ingest on the local machine. 
If the user has full flexibility, I imagine they'd be setting the folder name they want on their own machine - then they'd have to be careful with the docker volumes, right ? In your example /data/cwa-comic/ingest on my machine is not bound to anything on the container, unless I add it...  learning opportunity: can I use ${VAR}s in volume mapping in the docker compose yml? 
(sorry, rushing right now but will look into it later - ignore last push, it was with the "" defaults, I hadn't expected such a fast reply :D )

calibrain: So, for the first question, yes
You could manually edit the mounted folder in the docker-compose
But I think it's better to have it explicit, so people know exactly what they are doing and no surprises

Basically, when writing my docker compose, I know the exact path when putting my env variables and my volume mount

For the second question, yes, in general, you can have in your .env some VAR_NAME, and use it in the compose
But I personally don't like that as much, just because of my own deployment process I don't use ENV as much, but I probably should lol

Thank you so much for your work, and no worries, take your time !

TigreModerata: So ok, I had not read carefully (as per usual) about the different instances of CWA running... I focused in on my own needs and preferences (I don't even use CWA, I'm on the old desktop app 😆 ).

I've re-pushed, now with full flexibility.
My readme addition may be a bit redundant, I'm not sure what an average user knows.

TigreModerata:
Conflicts fixed, should be ok.

backend.py
```diff
@@ -206,7 +216,8 @@ def _download_book_with_cancellation(book_id: str, cancel_flag: Event) -> Option
             ws_manager.broadcast_status_update(queue_status())
         
         intermediate_path = INGEST_DIR / f"{book_id}.crdownload"
```
calibrain: Could you make the intermediate_path also the same _prepare_download_folder ?
This helps with some weird bugs of moving file cross NFS
Its easier to download at the same spot and just rename the extension

TigreModerata: I should have added/corrected the e2e testing file also, doing it now.

TigreModerata:
@calibrain I added some changes to the E2E_test.py, see comment in PR

testing/E2E_test.py
```diff
@@ -38,7 +40,7 @@ def check_download_status(book_id):
             continue
 
         # Check success conditions based on download_path
-        for status_key in ["available", "done"]:
+        for status_key in ["available", "done", "complete"]:
```
TigreModerata, line 42: Had to add "complete" to the options, I don't know why it never passed through "available" or "done". 
'available' never seems to get set, and 'done' seems to only be set if 'available' was set before... ?

calibrain:
Awesome, thank you
I merged something earlier, so I fixed the conflict
I will try this today or tomorrow and then merge, thank you so much ! 
And sorry for the back and forth

TigreModerata:
> Awesome, thank you I merged something earlier, so I fixed the conflict I will try this today or tomorrow and then merge, thank you so much ! And sorry for the back and forth

Back and forth was completely my fault, apologies all mine! 
Thanks for the app - I was just looking for somewhere to contribute, but this got me back into reading (after downloading all my favourite authors' works while testing 😄). I see light at the end of the doomscrolling tunnel 😆

testing/E2E_test.py
calibrain: no worries, I havent touched E2E_test in months, so I am amazed its still working

calibrain:
Hey ! I thought of it a bit more, and I am not sure about using enums for this
Its a bit janky and prone to failures, especially since we dont have a good track of type enforcing on this repo...
I sent you a PR : https://github.com/TigreModerata/calibre-web-automated-book-downloader/pull/1
to apply on top of this one if you feel thats okay

What do you think ?
The idea is to basically just use the dictionary of types -> folder directly, without going through the Enum intermediary

TigreModerata:
Yeah, makes sense - that Enum was from the first version, where I had fixed the names of respective folders, before they went to env vars, but now it really only complicated matters. 

Now it does look a lot more sensible!

You ended up having to do the whole thing yourself 🫤 (and thanks for being so kind about it!)

calibrain:
No worries ! I used Cursor to automate some of it so it was fairly quick !
I don't have permission to merge it on your fork, could you merge it and kick then update this PR, I will merge it here after
Thank you again for all great work !

calibrain:
Awesome, merged ! Thank you so much again !

PR #301, opened 2025-11-08
Web UI - Various UI / UX / QoL improvements

alexhb1:
Hi, these were a few core changes I've been using on my own install plus a bit of extra polish I've done for this PR. First time doing a PR but a big fan of this project and wanted to contribute, appreciate all you've done so far :)

---

**Main UI -** 

- Search-engine-like design on startup with the centered box. 
- The advanced options are available from the icon. The queue is hidden unless a download is active or complete, and then will remain visible like before. 

Screenshot 2025-11-08 at 16 14 03

---
**Downloading** - Responsive and dynamic buttons, and added notifications

Now when clicking download, the button will go through various states and display a spinning wheel. A small toast notification will also appear at the top right of the screen. 

1. Queuing download 
image

2. Added to queue (Button and notification)
image
image

3. Downloading (Button and notification)
image
image

4. Completed (Notification)
image

---

**Mobile UI improvements**

Improved space on mobile by tweaking the search cards like below. The desktop view stays the same as before. 

image

---
Some more misc things: 

- Tweaks to the details pane, including replacing the pop-up loading screen with an in-line button so it doesn't hijack your screen for ages 
image

- Added a calibre-web button. Add your URL as an env variable and it becomes an easy way to navigate back to your calibre library once you've finished downloading. This is only visible if the env variable is added. 
- The queue will auto-refresh every 5 seconds
- Cleaned up old UIKit CSS no longer used

### DIFF
--- a/static/css/styles.css
+++ b/static/css/styles.css
@@ -8,9 +8,6 @@
     --border-color: #e5e5e5;
     --loading-overlay: rgba(0, 0, 0, 0.5);
     --card-background: #fff;
-    --table-background: #fff;
-    --table-hover-background: #f8f8f8;
-    --table-border-color: #e5e5e5;
     --input-background: #fff;
     --heading-color: #333;
 
@@ -28,9 +25,6 @@
     --heading-color: #ffffff;
     --card-background: #1e1e1e;
     --border-color: #404040;
-    --table-background: #1e1e1e;
-    --table-hover-background: #2d2d2d;
-    --table-border-color: #404040;
     --input-background: #2d2d2d;
 }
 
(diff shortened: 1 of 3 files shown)


ZYancey:
Dang, I pull up github to start my own PR based on my own local FE rework and saw this 😆. GG, you beat me to it. I really like your design choices though. I think I will rebase my changes on top of yours and adopt your design. Then my PR will be a bit more localized to FE architecture changes like rebuilding on React/TS and setting up a component library in case the creator wants to implement a settings page or something instead of all of the env flags.

alexhb1:
Sorry to steal your thunder 🤣. Much appreciated. And I'm sure there's a lot more that can be done on the more technical side of the UI down the road too, I'm definitely more a design oriented guy anyway.

alexhb1:
Go see #302 instead!

PR #310, opened 2025-11-14
WebUI - Additional Search Features

alexhb1:
### Main Points / To Do List

- [X] New Compact mode, with automatic and manual activation
- [X] New List mode, additional manual view 
- [X] Move sorting options to the main search results pane - dropdown menu alongside view toggles 
- [X] New language handling, including default language and multi-select options. 
- [x] New details view 
- [X] Various refactoring, including reuseable components for the three search view components (Card, Compact & List), the download button, and a reuseable dropdown list component. 

---

### Card sizes: 

**Compact**
Screenshot 2025-11-15 at 15 35 55

Mobile: Compact by default:
Screenshot 2025-11-15 at 15 37 35

**List**
Screenshot 2025-11-15 at 15 35 24

Mobile: Optional
Screenshot 2025-11-15 at 15 37 59

---
### Redesigned details pane: 
Screenshot 2025-11-15 at 15 39 50

Mobile: 
Screenshot 2025-11-15 at 15 40 41

--- 
### Multi-select languages
Screenshot 2025-11-15 at 15 41 29


src/frontend/src/components/ResultsSection.tsx
```diff
@@ -16,34 +17,124 @@ export const ResultsSection = ({
...
 }: ResultsSectionProps) => {
+  const [viewMode, setViewMode] = useState<'card' | 'compact'>(() => {
+    const saved = localStorage.getItem('bookViewMode');
+    return (saved === 'card' || saved === 'compact') ? saved : 'card';
+  });
+
+  const [isMultiColumn, setIsMultiColumn] = useState(false);
+
+  useEffect(() => {
+    localStorage.setItem('bookViewMode', viewMode);
```
ZYancey, line 36: > Ability to save default sort and card view (Maybe - @ZYancey I think related to what you mentioned about local storage - is this something you could possibly explore?)

It looks like this is already implemented here, I initially assumed that this meant you wanted me to take a crack at implementing it, this looks good though and works as expected.

alexhb1: Sorry I've just recreated the branch and pulled in everything I had without realizing 😂. It works for the button state now yeah. The other question was mainly whether a config menu was a better option, setting defaults from startup etc. I think that was mentioned before.

ZYancey: I think there is room for both, because localstorage persists based on the browser being used. But perhaps we would want to set a default for all users or whatnot. This weekend I'll pull down your branch and set up the DB and see how easy it is to build a config page. Either way, I like this change

calibrain:
Currently, the front end seems to ignore the `LANGUAGE` env, and always defaults to english.
Also, selecting `ALL` does not seem to work

Could you add a Default which uses all languages selected in the env variable `LANGUAGE`
Also, make it that ALL just either removes the filter or adds all supported ones
Also, would it be possible to make the drop down menu multi selection ? If I want lets say English AND French in a single search ?

Thank you so much !

C.C. @ZYancey

alexhb1:
Will do :)

alexhb1:
@calibrain I've fixed the language selection default and added a multi-selecting list. Unfortunately I don't think "All" works the way we would like, I've checked that the app is sending zero language preference when "All" is ticked but the results still turn up mainly English, the other languages are too far down the search order to be displayed when sorting by "Relevant". This is the same on the AA website so I assume it's just how the data works. 

I think if anyone needs specific languages they'll have to use the dropdown or set the default env variable.

alexhb1:
All done. Can be safely merged all in one - this includes the login GUI changes too :)

calibrain:
Awesome work ! Thank you so much !

Issue #303, opened 2025-11-11
[BUG] Log spam: "Error parsing search result row"


Muppetteer:
## Bug Description
Every search generates multiple log errors because the AA website is including ads in between search result rows and the parser is not skipping those ad rows.

## Steps To Reproduce
1. Search for a book
2. Look at the logs

## Expected Behavior
Skip the ad rows and don't log an error

## Debug Information
`2025-11-10 10:52:09,308 - book_manager - ERROR - logger.py:16 - Error parsing search result row: list index out of range
Traceback (most recent call last):
  File "/app/book_manager.py", line 120, in _parse_search_result_row
    title=cells[1].find("span").next,
IndexError: list index out of range`

## Additional Context
This is a valid search results row:
`<tr class="group h-full even:bg-[#f2f2f2] hover:bg-yellow-100 aria-selected:bg-yellow-100 cursor-pointer relative "> ... </tr>`

This is an ad row:
`<tr class="[html.aa-logged-in_&amp;]:hidden"> ... </tr>`

So filtering the row search to only include rows with class="group" would seem to be an easy fix. I have not coded with BeautifulSoup before so this is me guessing after reading the docs but I think this will fix it:

Change book_manager.py line 91:
`for line_tr in tbody.find_all("tr"):`

to this:
`for line_tr in tbody.find_all("tr", class_="group"):`

Glennza1962:
same here

calibrain:
Its a bit harder than that, since depending on what is available, the UI seems to generate different tags and fields
Same as #300  and #307  I think
I am brianstorming on a reliable fix

calibrain:
Actually no, this is unrelated, sorry didnt properly read

Fixed in #312 
Currently only available in `:dev`, but will push to `:latest` in the upcoming week (we are testing some new UI changes in `:dev` right now)

PR #312, opened 2025-11-15
Skip ad rows in search result parsing

calibrain:
AA started injection an ad banner into their code, for now we start skipping this.
I am expecting later we will need to revisit this code for when they actually start injecting the add


calibrain:
Fixes #303

Issue #300, opened 2025-11-08
[BUG] Book title too long


tweety-tv:
## Bug Description
The name of the download file is not correctly fetch, it get's a long javascript code and try to use that as filename.
<img width="1301" height="242" alt="Image" src="https://github.com/user-attachments/assets/7a1ab5ad-40cc-47de-8952-2eefd193b297" />
When using USE_BOOK_TITLE the file can not be saved as the filename is too long (the javascript code)

It's not for every book, some work correctly. But for example https://annas-archive.org/slow_download/0618eb06a8f7cff3bc693f78e68ee4b9/0/3 doesn't work as expected

## Steps To Reproduce

1. Search books (ex Reece Barden, language Dutch)
2. Click "download" on any of the found books

## Expected Behavior
Parse book title correctly

## Debug Information
**Important**: Please enable debug mode before submitting a bug report by setting the environment variable:
```
DEBUG=true
```

When running in debug mode, a DEBUG button will appear in the interface. Please click this button to generate a debug zip file and attach it to this issue.

[cwa-book-downloader-debug_BUILD-2025-11-06-8ea2fee0bbdfdff225a313b95e495b5ba15faf6d_RELEASE-v0.2.4_20251108-075524.zip](https://github.com/user-attachments/files/23430625/cwa-book-downloader-debug_BUILD-2025-11-06-8ea2fee0bbdfdff225a313b95e495b5ba15faf6d_RELEASE-v0.2.4_20251108-075524.zip)

## Environment Information

- BASE OS: Ubuntu 22.04

calibrain:
Oh
It seems some book when missing information dont get rendered properly and the naming pipeline breaks
I will look into this, thank you for the report

calibrain:
This should be now fixed in `:dev`
The fix will be released in the next weeks to `:latest`
Thank you for the report !

Issue #307, opened 2025-11-14
[BUG] The downloaded file name is incorrect.


LemonPG:
## Bug Description
Some download filenames are incorrect.

## Steps To Reproduce
1. search"肖申克的救赎"
2. click first download

## Debug Information
**Important**: Please enable debug mode before submitting a bug report by setting the environment variable:
```
DEBUG=true
```

When running in debug mode, a DEBUG button will appear in the interface. Please click this button to generate a debug zip file and attach it to this issue.

[cwa-book-downloader-debug_BUILD-2025-11-06-8ea2fee0bbdfdff225a313b95e495b5ba15faf6d_RELEASE-v0.2.4_20251114-110710.zip](https://github.com/user-attachments/files/23538721/cwa-book-downloader-debug_BUILD-2025-11-06-8ea2fee0bbdfdff225a313b95e495b5ba15faf6d_RELEASE-v0.2.4_20251114-110710.zip)

## Environment Information

- BASE OS: [e.g. Windows 10, Ubuntu 22.04]
 XUbuntu 24.04

## docker-compose
services:
  calibre-web-automated-book-downloader-extbp:
    image: ghcr.io/calibrain/calibre-web-automated-book-downloader-extbp:latest
    environment:
      FLASK_PORT: 8084
      LOG_LEVEL: info
      BOOK_LANGUAGE: en
      USE_BOOK_TITLE: true
      TZ: America/New_York
      APP_ENV: prod
      UID: 1000
      GID: 100
      EXT_BYPASSER_URL: http://flaresolverr:8191
    ports:
      - 8084:8084
    restart: unless-stopped
    volumes:
      # This is where the books will be downloaded to, usually it would be
      # the same as whatever you gave in "calibre-web-automated"
      - /tmp/data/calibre-web/ingest:/cwa-book-ingest
      # This is the location of CWA's app.db, which contains authentication
      # details
      #- /cwa/config/path/app.db:/auth/app.db:ro
    HTTP_PROXY: http://127.0.0.1:7878
    HTTPS_PROXY: http://127.0.0.1:7878
    NO_PROXY: flaresolverr,127.0.0.1
    no_proxy: flaresolverr,127.0.0.1
  flaresolverr:
    image: ghcr.io/flaresolverr/flaresolverr:latest
  environment:
    HTTP_PROXY: http://127.0.0.1:7878
    HTTPS_PROXY: http://127.0.0.1:7878
[...]

PR #313, opened 2025-11-15
Improve book metadata handling in book_manager.py

calibrain:
Refactor book metadata extraction and add helper function.


calibrain:
Fix #300 and #307

PR #319, opened 2025-11-19
Fixes default authentication in docker-compose.yml

Sterahi:
The current docker compose does not allow for authentication & provides a less than clear error (`CWA_DB_PATH is set to /auth/app.db but this is not a valid path`).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,4 +27,4 @@ services:
       - /tmp/data/calibre-web/ingest:/cwa-book-ingest
       # This is the location of CWA's app.db, which contains authentication
       # details. Comment out to disable authentication
-      #- /cwa/config/path/app.db:/auth/app.db:ro
+      #- /cwa/config/path/app.db:/auth:ro


calibrain:
Could you share what is your current docker compose looks like ?

Sterahi:
@calibrain Here you go

```yml
services:
    image: ghcr.io/calibrain/calibre-web-automated-book-downloader:latest
    # Uncomment to build the image from the Dockerfile for local testing changes.
    # Remember to comment out the image line above.
    #build: .
    container_name: calibre-web-automated-book-downloader
    environment:
      FLASK_PORT: 8084
      LOG_LEVEL: info
      BOOK_LANGUAGE: en
      USE_BOOK_TITLE: true
      TZ: America/New_York
      APP_ENV: prod
      UID: 1000
      GID: 100
      CWA_DB_PATH: /auth/app.db  # Comment out to disable authentication
      # Queue management settings
      MAX_CONCURRENT_DOWNLOADS: 3
      DOWNLOAD_PROGRESS_UPDATE_INTERVAL: 5
    ports:
      - 8084:8084
    restart: unless-stopped
    volumes:
      # This is where the books will be downloaded to, usually it would be
      # the same as whatever you gave in "calibre-web-automated"
      - /tmp/data/calibre-web/ingest:/cwa-book-ingest
      # This is the location of CWA's app.db, which contains authentication
      # details. Comment out to disable authentication
      - /cwa/config/path/app.db:/auth:ro
```

It's the same as the default, but with the authentication enabled & with the one line change.

alexhb1:
Closing this, as far as I can tell the current compose setup does work fine for accessing the CWA app DB, there's no issue with access as far as my testing goes. If there's any error message I would assume it's a more broad issue with file system access. 

In any case, the app now offers built-in authentication options, so CWA isn't a hard requirement for this feature.

PR #331, opened 2025-12-08
Change APP_ENV from 'prod' to 'dev' default

calibrain:


alexhb1:
Hey, just seen the change. I'd probably advise not using dev all the time, there's some prod-specific functionality in the websockets using gevent instead of threading which helps with performance, same with enabling CORS in dev vs prod, which weakens security a little. 

Instead we could set app_env to prod by default, set HTTPS_COOKIE_SECURE to false by default and offer a flag in the docker compose to set this to true if needed for each user? Happy to PR with what I had in mind :)

calibrain:
Basically I expect if someone is using a reverse proxy, they would set the app to be PROD
If they are connecting manually through IP address or local DNS or localhost, DEV should be the default no ?

alexhb1:
Sure that makes sense. I guess I was just thinking if people are using local connection for daily use and some of the prod-specific features aren't enabled, wouldn't it be better if everyone used prod and it worked for everyone? The HTTPS / reverse proxy issue is only the HTTPS_COOKIE_SECURE flag, so would it be better to make that the option to enable when using reverse proxy instead of changing the entire environment? 

I just don't want to make it overcomplicated in terms of what versions people are using and when if that makes sense :)

calibrain:
Actually, you are right
APP_ENV was used mainly for deciding between using gunicorn or not
But I guess we can just remove APP_ENV and do that with the DEBUG flag instead

Sure, shoot me a PR when you have some time and ill merge that

alexhb1:
No problem, hope that all made sense. At least the dev by default change fixes the bug already so that bug is solved for now, and I'll have a look at removing APP_ENV and just using debug. Thanks man :)

PR #333, opened 2025-12-09
APP_ENV removal and secure cookie handling

alexhb1:
Hey, made the tweaks we discussed, plus a couple related fixes :)

- Removed APP_ENV entirely. All dev-specific functionality is enabled via `DEBUG: true` env var
- Set secure cookie handling to false by default, added to the readme to enable if exclusively using HTTPS connection
- Fixed healthcheck potentially not working with auth enabled
- Removed APP_ENV from docker compose files and made sure app.db lines are included in all versions. 

APP_ENV in people's existing composes should get ignored entirely and will be put on the default env, so no issues when updating.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,6 +14,7 @@ services:
       UID: 1000
       GID: 100
       # CWA_DB_PATH: /auth/app.db  # Uncomment to enable authentication (also uncomment volume below)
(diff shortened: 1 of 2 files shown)


calibrain:
Awesome work !
Thank you so much !
Sorry for the late review, I was about sick this week

PR #336, opened 2025-12-11
Download overhaul - DNS fallback, bypasser enhancements, revamped error handling, better frontend UX

alexhb1:
### 🌐 Network Resilience

- **Auto DNS rotation**: New `CUSTOM_DNS=auto` mode (now default) starts with system DNS and automatically rotates through Cloudflare, Google, Quad9, and OpenDNS when failures are detected. DNS results are cached to improve performance.
- **Mirror failover**: Anna's Archive requests automatically fail over between mirrors (.org, .se, .li) when one is unreachable
- **Round-robin source distribution**: Concurrent downloads are distributed across different AA partner servers to avoid rate limiting

### 📥 Download Reliability

- **Much more reliable downloads**: Improved parsing of Anna's Archive pages, smarter source prioritization, and better retry logic with exponential backoff
- **Download resume support**: Interrupted downloads can now resume from where they left off (if the server supports Range requests)
- **Cookie sharing**: Cloudflare bypass cookies are extracted and shared with subsequent requests, often avoiding the need for re-bypass entirely
- **Stall detection**: Downloads with no progress for 5 minutes are automatically cancelled and retried
- **Staggered concurrent downloads**: Small delays between starting concurrent downloads to avoid hitting rate limits
- **Source failure tracking**: After multiple failures from the same source type (e.g., Libgen), that source is temporarily skipped
- **Lazy welib loading**: Welib sources are fetched as a fallback only when primary sources fail (unless `PRIORITIZE_WELIB` is enabled)

### 🛡️ Cloudflare & Protection Bypass

- **DDOS-Guard support**: Internal bypasser now detects and handles DDOS-Guard challenges with dedicated bypass strategies
- **Cancellation support**: Bypass operations can now be cancelled mid-operation when user cancels a download
- **Smart warmup**: Chrome driver is pre-warmed when first client connects (controlled by `BYPASS_WARMUP_ON_CONNECT` env var) and shuts down after periods of inactivity

### 🔌 External Bypasser (FlareSolverr)

- **Improved resilience**: Retry with exponential backoff, mirror/DNS rotation on failure, and proper timeout handling
- **Cancellation support**: External bypasser operations respect cancellation flags

### 🖥️ Web UI Improvements

- **Simplified download status**: Removed intermediate states (bypassing, verifying, ingesting) — now just shows Queued → Resolving → Downloading → Complete
- **Status messages**: Downloads show detailed status like "Trying Anna's Archive (Server 3)" or "Server busy, trying next...", or live waitlist countdowns. 
- **Improved download sidebar**:
  - Downloads sorted by add time (newest first)
  - X button moved to top-right corner for better UX
  - Wave animation on in-progress items
  - Error messages shown directly on failed items
  - X button on completed/errored items clears them from the list

### ⚙️ Configuration Changes

- **`CUSTOM_DNS=auto`** is now the default (previously empty/system DNS)
- **`DOWNLOAD_PROGRESS_UPDATE_INTERVAL`** default changed from 5s to 1s for smoother progress
- **`BYPASS_WARMUP_ON_CONNECT`** (default: true) — warm up Chrome when first client connects

### 🐛 Bug Fixes

- **Download cancellation actually works**: Fixed issue where cancelling downloads didn't properly stop in-progress operations
- **WELIB prioritization**: Fixed `PRIORITIZE_WELIB` not being respected
- **File exists handling**: Downloads to same filename now get `_1`, `_2` suffix instead of overwriting
- **Empty search results**: "No books found" now returns empty list instead of throwing exception
- **Search unavailable error**: Network/mirror failures during search now return proper 503 error to client

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -1,6 +1,7 @@
 """Backend logic for the book download application."""
 
 import os
+import random
 import shutil
 import subprocess
 import threading
@@ -484,10 +485,16 @@ def concurrent_download_loop() -> None:
                 next_download = book_queue.get_next()
                 if not next_download:
                     break
-                    
+
+                # Stagger concurrent downloads to avoid rate limiting on shared download servers
+                # Only delay if other downloads are already active
+                if active_futures:
+                    stagger_delay = random.uniform(2, 5)
+                    logger.debug(f"Staggering download start by {stagger_delay:.1f}s")
+                    time.sleep(stagger_delay)
+
                 book_id, cancel_flag = next_download
-                logger.info(f"Starting concurrent download: {book_id}")
-                
+
                 # Submit download job to thread pool
                 future = executor.submit(_process_single_download, book_id, cancel_flag)
                 active_futures[future] = book_id
--- a/book_manager.py
+++ b/book_manager.py
@@ -243,11 +243,10 @@ def _parse_book_info_page(soup: BeautifulSoup, book_id: str) -> BookInfo:
     # 1. AA slow (no waitlist) - instant but can be slow
     # 2. Libgen - instant, external
     # 3. AA slow (waitlist) - has countdown timer but faster once started
-    # 4. Z-Library - last resort (hash lookup rarely finds results)
+    # Note: Z-Library disabled - download tokens are session-bound
     urls += slow_urls_no_waitlist if USE_CF_BYPASS else []
     urls += external_urls_libgen
     urls += slow_urls_with_waitlist if USE_CF_BYPASS else []
-    urls += external_urls_z_lib
 
     for i in range(len(urls)):
         urls[i] = downloader.get_absolute_url(network.get_aa_base_url(), urls[i])
@@ -281,7 +280,8 @@ def _parse_book_info_page(soup: BeautifulSoup, book_id: str) -> BookInfo:
             if format == "" and f.strip().lower() in SUPPORTED_FORMATS:
                 format = f.strip().lower()
             if size == "" and any(u in f.strip().lower() for u in ["mb", "kb", "gb"]):
-                size = f.strip().lower()
+                # Preserve original case but uppercase the unit (e.g., "5.2 mb" -> "5.2 MB")
+                size = re.sub(r'(kb|mb|gb|tb)', lambda m: m.group(1).upper(), f.strip(), flags=re.IGNORECASE)
             if content == "":
                 for ct in DOWNLOAD_PATHS.keys():
                     if ct in f.strip().lower():
@@ -293,7 +293,8 @@ def _parse_book_info_page(soup: BeautifulSoup, book_id: str) -> BookInfo:
                 if format == "" and stripped and " " not in stripped:
                     format = stripped
                 if size == "" and "." in stripped:
-                    size = stripped
+                    # Uppercase any size units
+                    size = re.sub(r'(kb|mb|gb|tb)', lambda m: m.group(1).upper(), f.strip(), flags=re.IGNORECASE)
     
     book_title = _find_in_divs(divs, "🔍")[0].strip("🔍").strip()
 
@@ -619,7 +620,8 @@ def download_book(book_info: BookInfo, book_path: Path, progress_callback: Optio
 
             logger.info("Resolved download URL [%s]: %s", source_label, download_url)
 
-            data = downloader.download_url(download_url, book_info.size or "", progress_callback, cancel_flag, selector, status_callback)
+            # Pass source page as referer (required by some sites)
+            data = downloader.download_url(download_url, book_info.size or "", progress_callback, cancel_flag, selector, status_callback, referer=link)
             if not data:
                 raise Exception("No data received from download")
             
--- a/cloudflare_bypasser.py
+++ b/cloudflare_bypasser.py
@@ -51,6 +51,110 @@ DISPLAY = {
 LAST_USED = None
 LOCKED = threading.Lock()
 
+# Flag to track if DNS rotated while a bypass was in progress
+# Chrome will be restarted after the current operation completes
+_dns_rotation_pending = False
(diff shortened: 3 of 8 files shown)


alexhb1:
@calibrain Related query / idea - How would you feel about making the extbp version the default and gradually move people over to it? In my own experience and doing a lot of network troubleshooting/testing for this PR, using flaresolverr was a much quicker and reliable experience overall.

No fault of the internal bypasser, but the cold startup and the time that takes just can't really be matched by flaresolverr always running in the background and and constantly maintained and updated for that purpose. Would also massively help with install size / resource usage and overall code complexity.

calibrain:
I love this idea !
I was already doing something similar to find domains that are working, but automatic dns would be great too !
By the way, does `USE_DOH` works for you ? It should bypass the ISP blocking 

> @calibrain Related query / idea - How would you feel about making the extbp version the default and gradually move people over to it? In my own experience and doing a lot of network troubleshooting/testing for this PR, using flaresolverr was a much quicker and reliable experience overall.
> 
> No fault of the internal bypasser, but the cold startup and the time that takes just can't really be matched by flaresolverr always running in the background and and constantly maintained and updated for that purpose. Would also massively help with install size / resource usage and overall code complexity.

Actually those are features, not bugs :P 
flaresolverr had issues in the past, such as memory leak, no logs, and really hard to debug
The internal bypasser code is actually very similar to what flaresolverr does, with CDP and all.
We could add an ENV for keeping the browser hot maybe ?
Right now it's cold start and clean up after 10min of unused browser
But we could easily start it on app start and keep it open in the background ? Something like "FAST_SEARCH : true" with a disclaimer that it uses more RAM

I don't mind telling people to use the extbp by default (we can update documentation for that), but I still want to keep the internal bypasser (it makes it so much easier to debug locally on my macbook when coding, I don't need to run a docker or start another server)

alexhb1:
Lol that's fair! I guess just from a user perspective it definitely seemed faster, but that makes sense to keep it so no issues there. 

I've made a tweak to deprioritise welib unless the env states otherwise, which saves an extra bypass run on most downloads if it already has a download from AA, which helps with speed. Adding a hot option might be good also

I didn't realise that about DOH though, might have a try on my network and see if I can add it into the auto behavior on this

--- 
@calibrain I've made auto mode use DOH by default. Works fine and probably helps in more situations, unfortunately Virgin Media is too strict for me 😭. 

Thinking a bit more, I still think extbp would work well as the default. You could add some UI notices so people could adjust their docker composes and then eventually move `:latest` onto extbp. Then, the internal bypasser remains available as an optional extra and dev tool. Entirely up to you though :)

alexhb1:
@calibrain everything should be functional now. I've done various bits of testing and haven't ran into any issues. AA fast / slow and welib all work fine for me and the DNS works correctly. To the end user there's barely any difference, except there's no config needed and downloads/searches work a whole lot more reliably. Definitely give it a test if you have chance, it might need a bit more time to try more scenarios to make sure nothing's borked that I've missed. 

(I think #333 could also be released as a hotfix too, the logout issue might be quite common if people are using auth+HTTP it's guaranteed to break)

calibrain:
Awesome, will try it out this weekend
And yeah, I finally got some time, tested the new changes and pushed it to latest (v0.3.2)

Thank you so much again for the awesome work !

alexhb1:
@calibrain  I'm now a huge fan of the internal bypasser 🤣. I've managed to add tons of custom features to it (Auto DNS and mirror switching, more custom recognition of AA websites and DDOS-guard walls, better error handling, and it seems to work with welib more reliably). 

I've also managed to add a "smart-warmup" option for pre-launching the bypasser when you open the Web UI. It'll spin up and wind down depending on if the Web UI is open or not, saving the need to load it up during active downloads. It'll also disable pre-loading if you're using the extbp, have CF_BYPASS disabled, or if you're using an AA donator key. Seems pretty elegant without taking up too many resources I think!

I've set it to be default on, but easily disabled in the compose.

calibrain:
Oh yeah, thats really smart ! 
Hook it to the heartbeat (/status page), since it gets automatically refreshed every few seconds, then have a 10-15min timer from last HB before killing the browser

If possible, could you separate the PRs ? (network stuff, browser stuff and warmup stuff)
If it's too much work, no worries, we can keep it in the same PR it's fine

alexhb1:
It uses the websockets implementation to detect whether the UI is active or not, which I guess would work similar to the heartbeat. Open the Web UI - > Websockets connect -> bypasses starts, and shuts down when it disconnects. If a download is active, it stays open until downloads are complete and then shuts down at the existing time delay. 

I think everything is pretty tied together so might be a struggle! Got a bit carried away 😂. I'll leave it for now because i'll probably break something if I try and separate into parts. Instead, I think it’ll just need some extra testing to make sure everything’s working okay before releasing. No rush on this one :)

alexhb1:
@calibrain I've been testing every type of scenario, source, connection type I can think of, and adding fixes as I go - that's the cause of all these little recent additions, but hopefully makes up for it in massively improved download reliability and speed :)

Biggest extra addition is _**saving cookies from cloudflare/ddos-guard after a first successful bypass**_ - all subsequent downloads for a period of time will use the saved cookies and successfully download with no bypass required! Especially valuable for multiple concurrent downloads, only a single bypass required and takes much shorter. And for any subsequent downloads, cookies stay valid for a period of time so they download instantly too.

calibrain (approved):

calibrain:
Wow, thats so many super cool features ! 
Nice job ! 
Thank you so much !

alexhb1:
No worries :). Appreciate it's a big bunch of changes but honestly it was quite fun to do, and hopefully contains a lot of fixes and improvements - hopefully less bug reports about downloads not working 😂

PR #373, opened 2025-12-29
fix: update auth priority and fallback logic

bischoffjeremy:
Reordered the auth priority because the old logic was misleading. It would automatically default to "builtin" mode if credentials existed, completely ignoring the CWA database even if you wanted to use it. You wouldn't even notice it was happening until you realized the DB integration wasn't actually active. This fix ensures explicit CWA auth takes priority so you don't have to wipe your settings just to switch methods.

Cheers,

Your swiss librarian ;)

### DIFF
--- a/cwa_book_downloader/main.py
+++ b/cwa_book_downloader/main.py
@@ -139,10 +139,9 @@ def get_auth_mode() -> str:
     """Determine which authentication mode is active.
 
     Priority: 
-    1. Explicit CWA
-    2. Built-in
-    3. Implicit CWA (fallback)
-    4. No auth required -> "none"
+    1. CWA (if enabled in settings and DB path exists)
+    2. Built-in credentials (if configured)
+    3. No auth required or error -> "none"
     """
     from cwa_book_downloader.core.settings_registry import load_config_file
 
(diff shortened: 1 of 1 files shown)


alexhb1:
Hey, thanks for the fix! Making sure CWA auth is respected is good (it's better than my initial code lol), but just looking through the steps I wonder if we can make this a little more robust?

Correct me if I'm wrong but it would seem like a fallback to CWA auth would occur if you have linked the DB file, but even if you haven't explicitly enabled CWA auth in settings? Could a simpler / clearer setup be ensuring that the CWA Auth option remains the true on/off switch for that option: 

1. If CWA_DB_PATH is set and CWA Auth is set to enabled -> Use CWA Auth
2. If builtin credentials are setup -> Use builtin
3. Else -> No Auth

This way there won't be any confusion about CWA auth inadvertently being used even if the user has set the toggle to off. Thanks :)

bischoffjeremy:
Absolutely, I thought the same thing! 
I was hesitant to change too much of the existing structure initially, which is why I left the fallback that was already there. 
But you're right, it's much better this way. 
In the new commit (ref 246a03a ), I've cleaned it up: 
strict priority for the CWA toggle first, then built-in, and I removed the implicit fallback entirely.

alexhb1:
Great stuff, thanks for both of these :)

PR #374, opened 2025-12-29
fix: default username to 'admin' if password is set but username is empty

bischoffjeremy:
Fixed an issue where hitting save with an empty username would fail silently. This led users to think their changes were saved when they actually weren't. Now it automatically defaults to "admin" if you set a password but leave the username blank, making the save process reliable.

Cheers,

Your swiss librarian ;)

### DIFF
--- a/cwa_book_downloader/config/security.py
+++ b/cwa_book_downloader/config/security.py
@@ -62,11 +62,15 @@ def _on_save_security(values: Dict[str, Any]) -> Dict[str, Any]:
     values.pop("BUILTIN_PASSWORD", None)
     values.pop("BUILTIN_PASSWORD_CONFIRM", None)
 
-    # Determine if we have a password (either new or existing)
-    has_password = False
-    
     # If password is provided, validate and hash it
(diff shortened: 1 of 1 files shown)


alexhb1:
Thanks for this! I'm not sure about silently setting a new username for the user without much awareness of how the code is behaving, instead could we use your username check to provide a notification to the user to retry and fill in the username field? 

E.g. for mismatched passwords we do this which provides an error popup on the UI: 

```
if password != password_confirm:
    return {
        "error": True,
        "message": "Passwords do not match",
        "values": values
    }
```

Could we do the same to explain that username is empty - allowing the user to correct it with a username of their choice? Thanks :)

bischoffjeremy:
Fixed the logic as suggested (re: 8d6a1e2) got rid of the silent 'admin' default and added a proper validation error instead, so users are forced to pick a username. 

Also, sorry for the Copilot noise. I've tried turning the reviews off but it's still commenting for some reason and I can't figure out how to kill it.

PR #394, opened 2026-01-05
Remove .org domain for AA and add alternative domains

calibrain:
Anna's Archive main domain, `.org` was seized.
Removing it from choice list and adding official alternatives
More details : https://torrentfreak.com/annas-archive-loses-org-domain-after-surprise-suspension/

### DIFF
--- a/cwa_book_downloader/config/settings.py
+++ b/cwa_book_downloader/config/settings.py
@@ -717,9 +717,10 @@ def download_source_settings():
             description="Primary Anna's Archive mirror to use. 'auto' selects automatically.",
             options=[
                 {"value": "auto", "label": "Auto (Recommended)"},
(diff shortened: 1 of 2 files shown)


calibrain:
@alexhb1 We might want in the future to read the alternatives from open-slum.org and use the most stable one ?

alexhb1:
I think something that grabs all the available domains from open slum and refreshes the app list on startup would be good.  

I’d probably keep the app auto rotation between mirrors just in case people have ISP specific blocks with one mirror vs another (I have it all the time lol)

alexhb1:
For the time being, I'll tweak the config so that all mirrors are user configurable in settings, you can add and choose a preferred option. It also means you can add new AA mirrors into the auto-rotation. Hopefully means no hard-coding every time a mirror fails.

PR #481, opened 2026-01-17
Remote path mappings, Client handling improvements

alexhb1:


shelfmark/release_sources/prowlarr/clients/rtorrent.py
```diff
@@ -142,6 +142,7 @@ def get_status(self, download_id: str) -> DownloadStatus:
                 "d.up.rate=",
                 "d.custom1=",
                 "d.complete=",
+                "d.up.total=",
```
veverkap, line 145: This is throwing an error now because there isn't a matching for this key.

2026-01-17 21:21:26,861 - shelfmark.release_sources.prowlarr.clients.rtorrent - ERROR - c - rTorrent get_status failed (ValueError): too many values to unpack (expected 8)

Issue #534, opened 2026-01-25
[BUG] RTorrent does not return path of downloaded files to Shelfmark


erkston:
## Bug Description
When downloading a book via RTorrent (rakshasa ver. 0.16.5), Shelfmark can not complete the import of downloaded files, instead gives "path not returned" error

## Steps To Reproduce
1. Configure Prowlarr with RTorrent ( v0.9.7+) as a client
2. Verify paths are configured correctly for Shelfmark and RTorrent
3. Search for and download a book from Shelfmark
4. Get error:

<img width="408" height="88" alt="Image" src="https://github.com/user-attachments/assets/3e9e705e-3518-426b-b4fb-864188e78f27" />

## Expected Behavior
Shelfmark to retrieve the path information from RTorrent correctly and complete the download import

## Debug Information

[shelfmark-debug_BUILD-2026-01-24-43e554b8aef7a11e0d41189102ed5fd745cddf44_RELEASE-v1.0.2_20260125-121255.zip](https://github.com/user-attachments/files/24847463/shelfmark-debug_BUILD-2026-01-24-43e554b8aef7a11e0d41189102ed5fd745cddf44_RELEASE-v1.0.2_20260125-121255.zip)

```
2026-01-25 12:12:15,789 - shelfmark.release_sources.prowlarr.clients.rtorrent - DEBUG - rtorrent.py:167 - Torrent data for 8F1C0E5CB405C3D85E29840E09CF13D5745EBFB8: ['8F1C0E5CB405C3D85E29840E09CF13D5745EBFB8', 1, 201514, 201514, 0, 0, 'shelfmark', 1]
2026-01-25 12:12:15,791 - shelfmark.release_sources.prowlarr.handler - DEBUG - handler.py:428 - Download 8f1c0e5cb405c3d85e29840e09cf13d5745ebfb8 complete, file_path=None
2026-01-25 12:12:15,794 - shelfmark.release_sources.prowlarr.handler - ERROR - handler.py:517 - Download client returned empty path for completed download. Client: rtorrent, ID: 8f1c0e5cb405c3d85e29840e09cf13d5745ebfb8. Check that the download client's completion folder is accessible to Shelfmark.
```

## Environment Information
- BASE OS: Ubuntu 24.04.3 LTS

## docker-compose
```
  shelfmark:
    image: ghcr.io/calibrain/shelfmark:latest
    container_name: shelfmark
    environment:
      - TZ=${TZ}
[...]

PR #549, opened 2026-01-27
fix newer versions of rtorrent

veverkap:
Closes https://github.com/calibrain/shelfmark/issues/534

This pull request enhances the rTorrent client testing and implementation by adding more robust checks for directory paths and improving how the base path is retrieved. The main focus is on verifying and obtaining the correct download and base directories for torrents.

### DIFF
--- a/scripts/test_clients.py
+++ b/scripts/test_clients.py
@@ -468,7 +468,7 @@ def test_rtorrent():
         torrent_list = client.d.multicall.filtered(
             "",
             "default",
-            f"equal=d.hash=,cat={torrent_id}",
+            f"equal={{d.hash=,cat={torrent_id}}}"
             "d.hash=",
             "d.state=",
             "d.completed_bytes=",


shelfmark/release_sources/prowlarr/clients/rtorrent.py
```diff
@@ -310,7 +310,14 @@ def _get_torrent_path(self, download_id: str) -> Optional[str]:
...
         """
         try:
-            base_path = self._rpc.d.get_base_path(download_id)
-            return base_path if base_path else None
+            details = self._rpc.d.multicall.filtered(
+                "",
+                "default",
+                f"equal=d.hash=,cat={download_id}",
+                "d.base_path=",
+            )
```
veverkap, line 320: @erkston was this the same that you did?

erkston: Yes `d.base_path` works for me testing xmlrpc calls directly. However I also found this snippet [in the wiki](https://github.com/rakshasa/rtorrent/wiki/Common-Tasks-in-rTorrent#getting-the-full-path-of-data-of-a-torrent):

> From 0.9.2, d.base_path attribute can be empty sometimes (e.g. when a download is stopped). The following method always returns the full path of data of a download.
>```# gets the full path of data of a torrent (it's a workaround for the possibly empty 'd.base_path' attribute)
>method.insert = d.data_path, simple, "if=(d.is_multi_file), (cat,(d.directory),/), (cat,(d.directory),/,(d.name))"
>```

`d.base_path` was never blank in my testing (tested single and multi file torrents both seeding and stopped) but I did notice it returns different things for single file and multi file torrents. If there's only a single file it returns the full path including the filename (e.g. /downloads/EverybodyPoops.epub), while for multi file torrents whose files are in a directory it will return that directory (e.g. /downloads/Everybody-Poops-Complete-Series).

I'm not sure if that makes a difference but thought it worth mentioning

alexhb1:
Looks good, thanks :)

greywolf1499:
Hi team, I appreciate all the great work you guys have done on this project. I know this issue has been closed but I'm running into an error, I believe due to this PR, where RPC2 calls to rtorrent 0.9.6 fail because of the `d.multicall.filtered()` method being undefined. I wanted to confirm if this is intentional, and if not, could backward compatibility be added? As for the use case, I'm using a seedbox and while I have raised a ticket with the provider querying this, I don't have an option to upgrade this on my end. All the *arr containers appear to have this sort of backward compatibility which is why I'm seeing this error for the first time.

PR #557, opened 2026-01-29
[DOCS] Add more tools confirmed working

webysther:
Confirmed working with automatic importing in calibre and calibre-web.


alexhb1:
Thanks

Issue #559, opened 2026-01-30
Feature Request: Enhanced Path Template Variables with Arbitrary Prefix/Suffix


seekermarcel:
## Problem

Currently, path templates only support a limited set of prefix/suffix characters around variables (spaces, dashes, dots, underscores, slashes, brackets). This prevents users from creating templates like `{Vol. SeriesPosition - }` where "Vol." is a meaningful prefix that should only appear when the variable has a value.

**Current behavior:**
- Template `{Vol. SeriesPosition - }` fails to parse because "Vol." contains letters
- Users cannot add descriptive text around conditional variables
- Workarounds require awkward compromises in naming conventions

## Use Case

When organizing a book series, I want filenames like:

| SeriesPosition | Desired Output |
|----------------|----------------|
| 2 | `Vol. 2 - The Way of Kings.epub` |
| (empty) | `The Way of Kings.epub` |

The prefix "Vol." and suffix " - " should only appear when `SeriesPosition` has a value. If the book isn't part of a series (no position), the entire block should collapse to nothing.

Other examples:
- `{Book SeriesPosition of the Series}` → "Book 2 of the Series" or ""
- `{Part PartNumber}` → "Part 05" or ""
- `{(Year)}` → "(2020)" or ""

## Proposed Solution

Allow arbitrary text as prefix/suffix inside `{...}` blocks. The system detects known variable names within the block and treats everything before/after as conditional prefix/suffix.

seekermarcel:
I'm a developer myself and already build this. I'll link the PR. Please tell me if my style is fine for you or if something should be changed <3

PR #560, opened 2026-01-30
Enhance naming templates with arbitrary prefix/suffix support

seekermarcel:
As described in https://github.com/calibrain/shelfmark/issues/559
I would like the option to use prefix/suffix text as part of my file handling.

I appreciate every feedack


alexhb1:
Looks good! Seems to work fine on my end. Thanks

PR #569, opened 2026-02-02
chore: minor README update

solsetcassette:
Small change to readme file


ronnoceel:
Of note (in case anyone wondered like I did): prowlarr does not support audiobookbay.

marcSoda:
Is the ABB integration cancelled?

marcSoda:
Why was this PR abandoned? It looks like significant work was done to support ABB integration. ABB support for Shelfmark would be HUGE. Is it no longer in the works?

Issue #571, opened 2026-02-02
[BUG] Base Path does not get applied to image URLs


andykelk:
## Bug Description
When a "Base Path" is specified under Settings -> Advanced, images fail to display

## Steps To Reproduce

1. Set a base path in the advanced settings (e.g. `/shelfmark`)
2. Visit the home page (e.g. `https://my.domain/shelfmark`)
3. Inspect the logo image, the URL does not include the base path (e.g. `https://my.domain/logo.png`) and leads to a 404

## Expected Behavior

The image should use the base path (e.g. `https://my.domain/shelfmark/logo.png`)

## Debug Information
**Important**: Please enable debug mode before submitting a bug report by setting the environment variable:
```
DEBUG=true
```

debug: [shelfmark-debug_BUILD-2026-01-26-10bfaec793b2dad0cf594379150b215eb3fa6f0d_RELEASE-v1.0.3_20260202-102743.zip](https://github.com/user-attachments/files/25008068/shelfmark-debug_BUILD-2026-01-26-10bfaec793b2dad0cf594379150b215eb3fa6f0d_RELEASE-v1.0.3_20260202-102743.zip)

## Environment Information

- BASE OS: Ubuntu 24.04

PR #572, opened 2026-02-02
Fix base path resolution timing issue for subpath deployments

andykelk:
When deployed under a URL prefix (e.g., /shelfmark), images loaded by React were not respecting the base path, causing 404 errors. The logo would incorrectly load from /logo.png instead of /shelfmark/logo.png.

The root cause was that the BASE_PATH constant was being initialized at module load time, before the DOM was fully parsed. This meant document.querySelector('base') returned null, causing BASE_PATH to default to '/' regardless of the actual base tag value.

Changed to lazy initialization pattern where the base path is resolved on first access, ensuring the DOM and base tag are ready.

Fixes [#571](https://github.com/calibrain/shelfmark/issues/571)

### DIFF
--- a/src/frontend/src/utils/basePath.ts
+++ b/src/frontend/src/utils/basePath.ts
@@ -30,16 +30,23 @@ const resolveBasePath = (): string => {
   }
 };
 
-const BASE_PATH = normalizeBasePath(resolveBasePath());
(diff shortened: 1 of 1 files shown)


alexhb1:
Looks good, thanks!

Issue #659, opened 2026-02-25
[BUG] Search state not reset when switching media types + searching again


spin-drift:
To reproduce:

1. Search for an ebook or audiobook.
2. Download something from the list.
3. Without changing search terms, switch media type (from ebook to audiobook or vice-versa).
4. Hit Search again.

The results won't change. And when you click into the details for a book, you'll see **Downloaded** where you would otherwise see **Find Downloads**. Clicking on **Downloaded** does nothing.

If you refresh, you'll be able to proceed without a problem.

Issue #658, opened 2026-02-25
[BUG] Book on MAM, but not in any release list


spin-drift:
I hate to bring things to the level of individual books, but in case it's instructive: The book "The Long Game" by Rachel Reid does not show up in any possible combination of search terms. Searching by author shows all books but that one. ISBN, nothing. Exact title, long list of books, but not that one. When using author name + exact book title, it simply returns no results. Very curious.

It is already seeding in my client, but so are lots of other books I've successfully added, so I'm not inclined to think it's that.

Issue #657, opened 2026-02-25
[BUG] "Test connection" shouldn't reset config flow


spin-drift:
In any config area that involves a "Test" button, clicking it will wipe out and reset the configuration state unless the user has already clicked Save.

Issue #654, opened 2026-02-25
[Matching] Assume book without number is #1


spin-drift:
Example: Selecting **Find Downloads** from `He Who Fights With Monsters` has book #4 as the top/default, with the actual book (without a number) much further down the list.

I'm not sure how Shelfmark's download ranking system works, but in this case, I'd expect it to know that the title + ANY number is not what we're looking for—or, if so, it should be #1.

I can see this getting complicated in cases where a book's exact title is also #X in a series, and a tracker release mentions that. In those cases, I might be looking for things like:

Book Title: {Anything else} #Number ← #Number probably refers to series position.

PR #660, opened 2026-02-25
Patch: Various fixes

alexhb1:
Various fixes from the last couple days: 

- Add manual approval option for book/audiobook requests (#651)
- Add flagged HTTP headers 
- Add filesystem fallback - copy + delete when hardlink/move fails across filesystems (#647)
- Dependency updates
- Tweak frontend test config (simplified tsconfig for tests)
- Fix overlapping sort scoring in release modal - duplicate scoring keys caused incorrect release ordering (#654 )
- Fix stale search session after download - search state was not refreshed when returning from a download (#659)
- Fix multi-format release filtering - releases with multiple formats were incorrectly excluded by the format filter (#658)
- Fix config persistence when action button is used - clicking "Test connection" reset unsaved settings (#657)
- Fix request grid text positioning in admin request policy panel 
- Fix rTorrent path discovery (#653)
- Fix `/login` API check (#650)

### DIFF
--- a/src/frontend/src/components/ReleaseModal.tsx
+++ b/src/frontend/src/components/ReleaseModal.tsx
@@ -267,9 +267,7 @@ function getTitleMatchScore(title: string, titleCandidate: string): number {
 
   if (normalizedTitle.startsWith(titleCandidate)) {
     score += 6000;
-  }
-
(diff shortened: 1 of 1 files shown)


spin-drift:
💜💜💜

PR #717, opened 2026-03-07
Wishlist function

blades:
https://github.com/calibrain/shelfmark/discussions/551

This requested a wishlist function, and I've seen several requests for this and would find it useful myself.
I'm not the best at React, but I've knocked up something that I'm using locally that I figured other people (particularly those who want this functionality) might like to have a look at.

If there's a better approach to take, or for some reason this function *really* isn't welcome, I get it.

I'm running this on my local install and it seems to be working, but I'm in the early days of testing. I've added unit tests to cover it, and really it's not massively complex or smart - I've tried to fit cleanly into the existing codebase as much as possible, but I'm not a React coder by any stretch and there's probably a better/cleaner way to do this. I took the approach that the list is as dumb as possible, and it's only used as a way to easily repeat a search.

I've also not paginated this to mimic the existing activity sidebar's functionality, but if someone's got a ton of wishlist items that might be required in the future.


blades:
Also added lazy loading to the images; in theory there's a scroll bar for a long list, so unless people have stupidly long wishlists then this should be sufficient.

alexhb1:
Hi, appreciate the work on this, but this falls a bit outside the scope of the project going forward - particularly adding a new user workflow which I want to avoid. For lists we have Hardcover list integration that covers this off quite well without adding book-state type features into the app itself.

I think an efficient alternative might be the option to add books to a Hardcover list via the details UI, I'll see if that's a nice quick addition.

blades:
So the problem there is that once the list is in hardcover, there's no way to go from that list to repeat the search. The only list function on hardcover that's available would be something like 'want to read' which serves a very different purpose - a book on that list might already exist in your library but not have been read yet, as opposed to a book that you haven't yet been able to find.

This has been requested multiple times because people (myself included) want to rerun the search quickly and conveniently without manually having to type in every book. The list is merely the mechanism for storing the list of searches.

If I've missed something in hardcover or if there's a better way to achieve the ability to save and re-run searches let me know and I'll see if I can figure a way to do that instead, especially given it's been requested multiple times in the last week alone that I've seen. Alternatively, if there's an implementation of this that better fits your vision for the project let me know and I'll work on that instead.

alexhb1:
This already exists in the Hardcover implementation. E.g. select your list and view all books to download as you like: 

Screenshot 2026-03-07 at 09 39 45
Screenshot 2026-03-07 at 09 40 13

The only missing piece is somewhere to add a search result to your Hardcover list, or create a new one, to close the loop. It's a simple tweak, I'll have a look at adding that for the next update.

blades:
I had no idea that even existed - yeah, would definitely make more sense to use someone else's functions for list management rather than recreating it here.
Cool, thanks for the info - I'll have a look and see if I can figure that out instead, and will post up a fresh PR if I can get a handle on it.

alexhb1:
Not to worry, I've been working on it a little this afternoon anyway, I'll merge in a couple hours for testing. 

Search results: 
Screenshot 2026-03-07 at 14 32 06

Details modal: 
Screenshot 2026-03-07 at 14 32 32

Release modal: 
Screenshot 2026-03-07 at 14 32 45

blades:
It looks like you're already well ahead on that feature, so I'll back away and just say thank you at this point :)

PR #778, opened 2026-03-18
Remove audible provider

alexhb1:


joeycsare:
Hi, sorry why was this removed again? Is there a more general way to add providers now?

PR #865, opened 2026-04-11
Add optional CWA sidecar manifests for folder deliveries

EdLaughton:
## Summary

This adds an optional `.cwa.json` provenance sidecar for Shelfmark folder deliveries so Calibre-Web-Automated can ingest exact Hardcover identifiers alongside the delivered file.

## Why

CWA can use exact Hardcover identifiers during ingest instead of relying on fuzzy title/author matching.

## What changed

- added CWA sidecar manifest generation for folder deliveries using exact trusted Hardcover provenance
- added `ENABLE_CWA_SIDECAR_MANIFEST` to Shelfmark's normal downloads settings/config model
- defaulted sidecar emission to disabled for production-safe behavior
- gated folder-output sidecar emission behind that setting while leaving ordinary delivery behavior unchanged
- documented the setting, filename pattern, payload shape, and integration boundary
- added focused tests for enabled vs disabled behavior, default behavior, filename/content, and weak provenance handling

## Config and behavior

- setting: `ENABLE_CWA_SIDECAR_MANIFEST`
- default: `false`
- output filename: `<delivered filename>.cwa.json`
- emission is additive only and does not change ordinary downloads when disabled
- when enabled, Shelfmark only emits the sidecar when exact trusted Hardcover provenance is available
- weak or untrusted provenance does not produce invented identifiers

## Design boundary

Shelfmark emits the sidecar when enabled. CWA consumes it during ingest. Ordinary Shelfmark delivery still works without it.

This PR does not add:

- auto-download policy changes
- release-selection UI changes
- source/provider preference UI changes
- CWA-side request impersonation or request attribution changes

## Validation

```bash
python -m pytest -o addopts=''   tests/config/test_download_settings.py   tests/download/test_cwa_sidecar.py   tests/download/test_orchestrator_status.py   tests/download/test_orchestrator_retry.py   tests/download/test_orchestrator_lifecycle.py   tests/core/test_request_routes_api.py   tests/core/test_requests_service.py   tests/core/test_download_processing.py -q
```


alexhb1:
Hi, appreciate the work but I'll have to decline this for now and pick it up later. A couple things: 
- I'm already planning a similar feature but with a different level of scope. Ideally I don't want to limit such features to a single tool (CWA) and a single metadata provider (Hardcover). So I think a more generic implementation makes sense 
- I'm helping out on the Grimmory project and the book formatting, sidecar, metadata side of things is still in flux, so I'm waiting until that's locked down to do anything metadata related. 

Also, this PR is based on an old branch so it undoes a ton of linter fixes, could you rebase on main rather than merge? 

Happy to pick this up later on, the actual sidecar writer looks solid, but likely will need to be adapted to a more generic approach via the DownloadTask model so it works with all search sources, and the ability to output into various JSON metadata shapes (CWA, audiobookshelf, Grimmory etc). Thanks

EdLaughton:
> Hi, appreciate the work but I'll have to decline this for now and pick it up later. A couple things:
> 
> * I'm already planning a similar feature but with a different level of scope. Ideally I don't want to limit such features to a single tool (CWA) and a single metadata provider (Hardcover). So I think a more generic implementation makes sense
> 
> * I'm helping out on the Grimmory project and the book formatting, sidecar, metadata side of things is still in flux, so I'm waiting until that's locked down to do anything metadata related.
> 
> 
> Also, this PR is based on an old branch so it undoes a ton of linter fixes, could you rebase on main rather than merge?
> 
> Happy to pick this up later on, the actual sidecar writer looks solid, but likely will need to be adapted to a more generic approach via the DownloadTask model so it works with all search sources, and the ability to output into various JSON metadata shapes (CWA, audiobookshelf, Grimmory etc). Thanks

Thanks, that all makes sense.

I’m happy to pause this for now and pick it up again once the metadata/sidecar direction is more settled.

Understood as well on the broader scope: if/when this comes back, I agree it should likely sit at a more generic level via the DownloadTask flow rather than being tied specifically to CWA/Hardcover, with support for different output shapes.

I’ll also make sure any future version is rebased cleanly onto current `main` rather than merged from the older branch base.

Appreciate the review — glad the sidecar writer itself looks like a reasonable starting point, and I’m happy to revisit/adapt it later when the surrounding model is clearer.

PR #867, opened 2026-04-11
Newznab capability

blades:
I've added a plugin using the same architecture as the prowlarr plugin to enable Newznab as a source.
I've tested locally with nzbhydra2 and it all seems to work as intended.
I've added some unit tests for this feature, and found that a couple of other unit tests weren't behaving so fixed those up while I was at it. I also ran all of the linters in the makefile against it and fixed those up, too, so hopefully this should be as clean and as compatible as it can be.


alexhb1:
Hi, looks great. Works well for me. 

You've caught me at a slightly awkward time, there's been a ton of lint + quality + test changes these last couple days so things might be a little out of date. Could you rebase with just `/release-sources/newznab` and `/tests/newznab/` additions included and make sure it's still passing with the new checks? Then should be good to go, thanks :)

blades:
Sure thing - this should now be updated. Nice to see the additional linting in the makefile as well - probably the most comprehensive suite I've seen, tbh.
Thanks - glad it's a useful addition :)

PR #897, opened 2026-04-17
feat: adding support for output to litara

deranjer:
Following booklore's setup, this adds support for litara ereader upload to book drop. Unlike booklore litara does not support uploading directly to libraries, it must go through the book drop process. This allows shelfmark to submit a file directly to litara's book drop, which allows the admin to approve the book and that adds it to the library.

EDIT: Ah, should have looked at the Makefile closer. Let me run all the required tests, add a test file for litara, then update this PR.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -42,7 +42,7 @@ services:
 
 Notes:
 - Point `/books` to your library ingest folder (Calibre-Web, Booklore, Audiobookshelf, etc) for automatic import.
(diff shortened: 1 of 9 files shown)


alexhb1:
Hi, appreciate the work, but given Litara looks very new i'd prefer to avoid the precedent of adding too many custom output options immediately when a new app is launched. I presume Litara has a bookdrop folder that would work with shelfmark's folder output, no API required?

deranjer:
Yes, it has a bookdrop folder.

PR #916, opened 2026-04-25
Remove immutable to allow wal checkpoints

### LINKED ISSUE #915: [BUG] Stale password hashes when used with CW auth reads due to immutable=1 flag
## Bug Description

CW/CWA use wal mode by default, which requires checkpoints to update the data to the latest.

By setting immutable=1, checkpoints won't run and your reading of sqlite data will have stale values

## Steps To Reproduce
1. Run CWA/CW
2. Change Admin password from admin/admin123 to anything else
3. Run shelfmark with CWA db auth
4. Shelfmark fails to login, because the checkpoint has not been run and has old data

## Expected Behavior
Shelfmark should read the current committed state of app.db, including any data in the WAL

Suggested fix: Drop immutable=1. mode=ro alone is sufficient for read-only access and respects WAL. Or document that users must run PRAGMA wal_checkpoint(TRUNCATE) after CWA changes.

humaidk2:
Fixes https://github.com/calibrain/shelfmark/issues/915

## Description

Remove immutable=1 to allow wal checkpoints to fire and get latest auth info.

Note db is still in readonly mode, thanks to mode=ro

## Test case

1. Run CWA/CW
2. Change Admin password from admin/admin123 to anything else
3. Run shelfmark with CWA db auth
4. Shelfmark should be able to login after password change

### DIFF
--- a/shelfmark/core/admin_routes.py
+++ b/shelfmark/core/admin_routes.py
@@ -134,7 +134,7 @@ def _sync_all_cwa_users(user_db: UserDB) -> dict[str, int]:
         raise FileNotFoundError(msg)
 
     db_path = os.fspath(CWA_DB_PATH)
-    db_uri = f"file:{db_path}?mode=ro&immutable=1"
+    db_uri = f"file:{db_path}?mode=ro"
     conn = sqlite3.connect(db_uri, uri=True)
     try:
         cur = conn.cursor()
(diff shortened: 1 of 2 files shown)


alexhb1:
Thanks!

alexhb1:
Hi, after some testing and reports like #926, i'm gonna revert this one and add a note in settings to restart shelfmark after changing your password externally.

humaidk2:
Sorry for the bug, I did retest it and works fine for me. Yeah, if you don't make the the mount writeable and add the Wal, shm files mounted, It's not gonna work correctly.Totally understandable to just add a note to restart both when changing password.

Personally, saw that you have oidc support, so gonna switch to that. Appreciate all the hard work in the project.

PR #1058, opened 2026-06-14
Add `content_type=combined` URL parameter and `FORCE_COMBINED_SEARCH` user setting

spin-drift:
Adds support for `content_type=combined` in URL search parameters, letting users force combined-mode searches via a bookmarkable link rather than relying on the last-used preference from localStorage.

The override is applied only in Universal mode and only when combined mode is actually available (universal enabled, `show_combined_selector` on, neither content type blocked by policy). Otherwise, it's silently ignored, consistent with how `content_type=ebook`/`audiobook` already behave outside Universal.

Existing `content_type=ebook`/`audiobook` URLs now also force combined mode off, so the URL is authoritative regardless of prior preference. 

Also adds a per-user `FORCE_COMBINED_SEARCH` setting that locks combined mode on whenever it's available.

URL `content_type=ebook`/`audiobook` overrides are also superseded by force-combined for the same reason: the search bar wouldn't let users switch back, so honoring the URL param would leave them in a state they couldn't escape from.

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -275,7 +275,7 @@ Show the option to search for and download both a book and audiobook together.
 
 **Always Use Combined Search**
 
-Force combined search whenever it's available. Locks the combined toggle on; the lock icon replaces the chain in the search bar.
+Force combined search whenever it's available. Locks the combined toggle on.
 
 - **Type:** boolean
 - **Default:** `false`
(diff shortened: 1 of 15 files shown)


calibrain (approved):

calibrain:
Thank you very much for your contribution !

Issue #795, opened 2026-03-23
[BUG] Shelfmark not passing prowlarr seed limits to torrent client


adman234:
## Bug Description

As I understand, the feature to pass seeding limits (time/ratio) from the indexer to the torrent client should be supported by Shelfmark with the latest updates, but I don't seem to have this option in the settings and it doesn't seem to be working.

## Steps To Reproduce

1. Using shelfmark, prowlarr qbittorrent in docker containers.
2. My other apps (sonarr, radarr) are able to pass seeding limits to my torrent client
3. Shelfmark doesn't seem to follow the preferences set in prowlarr

## Environment Information

- BASE OS: Unraid, running shelfmark in Docker container

jdwpom:
In your download client settings, do you have your torrent completion action set to 'keep' or 'remove'? I suspect the latter doesn't respect Prowlarr, while the 'keep' setting does, but then doesn't remove the torrent after your seed time.

adman234:
> In your download client settings, do you have your torrent completion action set to 'keep' or 'remove'? I suspect the latter doesn't respect Prowlarr, while the 'keep' setting does, but then doesn't remove the torrent after your seed time.

I currently have it on "keep". I just tried the "remove" setting which states:

"Remove deletes the torrent from your client immediately after import (stops seeding, files are kept)"

I tested and it doesn't respect the prowlarr seed time. I looked through the settings and didn't see anything else related to this setting, so I'm not sure how to get it to work.

Wargiboy:
I also have been unable to get this feature working correctly, seed times work well for me with radarr and sonarr but shelfmark still does not seem to pass them through from prowlarr to qbittorrent

adman234:
Unfortunately, I am still not able to get this feature working with the newest version. Seedtime in qbittorrent always just shows infinity in the UI. But when I right click to adjust, I am seeing a seedtime of 259,200 minutes, which suggests maybe there is a unit mismatch here? It's trying to pass through the default seedtime for this indexer of 3 days but passing it through as 3 * 60 days instead. And it's not respecting my input of 7200 minutes and using the default instead.

I am using MaM, qbittorrent, and shelfmark in unraid docker containers. Let me know if any logs or other info is needed.

adman234:
I saw another PR for this issue, and I'm now testing the :dev container but still having issues.

[...]

PR #1110, opened 2026-07-09
fix(prowlarr): prevent silent loss of indexer seed limits on transient Prowlarr API failures

adman234:
# fix(prowlarr): prevent silent loss of indexer seed limits

Related to #795, though not a fix for that specific (closed) report — see note below.

## Problem

With "Use Prowlarr seed preferences" enabled, ~5–10% of torrent grabs are added to the download client without their configured share limits and seed indefinitely (∞ ETA in qBittorrent).

Seed limits are resolved once at search time. `get_indexer_seed_settings()` builds on `get_indexers()`, which swallows all API errors and returns `[]`. A transient failure of the `/api/v1/indexer` call therefore produces an empty settings dict that is indistinguishable from "no limits configured", while the search itself (separate HTTP calls) still succeeds. Every result from that search is cached without `configuredSeedTimeMinutes`; grabbing one sends the torrent to the client with no limits.

Compounding factors: `cache_release()` is last-write-wins by GUID, so one degraded search can strip enrichment from a previously good cache entry; and the retry fields persisted at queue time snapshot the same missing values, so retries reproduce the failure.

## Changes

- **`api.py`** — `get_indexers()` / `get_enabled_indexers_detailed()` gain a keyword-only `raise_on_error` (default `False`, existing behavior unchanged). `get_indexer_seed_settings()` uses it, so fetch failures now propagate and an empty dict strictly means "nothing configured".
- **`source.py`** — searches fetch settings via `_fetch_indexer_seed_settings()`, which maintains a module-level last-known-good copy (merged on each success) and falls back to it with a warning when the fetch fails. After one successful fetch, results can no longer be cached un-enriched.
- **`handler.py`** — grab-time safety net in `_resolve_download()`: if seed preferences are enabled, the release is a torrent, and no configured limits are present in the cached result, the handler re-resolves the limits from Prowlarr for that indexer (`restrict_to=[indexerId]`) before adding to the client. If limits still can't be resolved, a warning is logged so the condition is visible instead of silent.
- **Tests** — regression coverage: last-known-good fallback (success updates cache, failure falls back, failure with no history returns empty, fallback copy is mutation-safe) and grab-time fallback (used when enrichment is missing, tolerates Prowlarr being down, skipped when enrichment is present). Existing test stubs for `get_enabled_indexers_detailed` updated to accept the new kwarg.

## Testing

- `uv run pytest tests/prowlarr/test_handler.py tests/prowlarr/test_source.py tests/prowlarr/test_integration_handler.py` — 93 passed, 2 skipped (Python 3.14.4)
- Full `tests/prowlarr` run has 11 pre-existing failures on this environment (Windows path-separator assertions in the qBittorrent/NZBGet/SABnzbd/Transmission client tests, e.g. `/downloads/x` vs `\downloads\x`); confirmed these also fail on unpatched `main` and are unrelated to this change
- `uv run ruff check` / `ruff format --check` — clean on touched files
- `uv run basedpyright` — 0 errors on touched files

No behavior change when `PROWLARR_USE_SEED_PREFERENCES` is disabled; the fallback path only activates when the preference is on and enrichment is missing for a torrent.

---

**Note on #795:** this PR references #795 for background context on the seed-limits feature, but it does not fix that report — #795 was about seed limits not being converted/applied at all (a units mismatch), and was already fixed by #946 / #959. This PR fixes a separate, still-present bug: `get_indexers()` silently swallowing transient API errors, which intermittently drops seed limits even when the feature is otherwise working correctly.

### DIFF
--- a/shelfmark/release_sources/prowlarr/api.py
+++ b/shelfmark/release_sources/prowlarr/api.py
@@ -158,24 +158,35 @@ class ProwlarrClient:
             logger.info("Prowlarr connection successful: version %s", version)
             return True, f"Connected to Prowlarr {version}"
 
-    def get_indexers(self) -> list[dict[str, Any]]:
-        """Get all configured indexers."""
+    def get_indexers(self, *, raise_on_error: bool = False) -> list[dict[str, Any]]:
+        """Get all configured indexers.
+
+        Args:
+            raise_on_error: When True, propagate API failures instead of
+                returning an empty list. Callers that must distinguish
+                "no indexers" from "the request failed" should set this.
+
+        """
         try:
             return _normalize_json_object_list(
                 self._request("GET", "/api/v1/indexer"),
                 context="Prowlarr indexer list",
             )
         except _PROWLARR_CLIENT_ERRORS:
+            if raise_on_error:
+                raise
             logger.exception("Failed to get indexers")
             return []
 
-    def get_enabled_indexers_detailed(self) -> list[dict[str, Any]]:
+    def get_enabled_indexers_detailed(
+        self, *, raise_on_error: bool = False
+    ) -> list[dict[str, Any]]:
         """Get enabled indexers, including implementation metadata.
 
         Note: Prowlarr indexer "name" is user-configurable; prefer
         "implementation"/"implementationName" for stable identification.
         """
-        indexers = self.get_indexers()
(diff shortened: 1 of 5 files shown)


calibrain:
Thank you for the contribution !

PR #1123, opened 2026-07-19
fix(naming): respect subfolder order when assigning audiobook part numbers

### LINKED ISSUE #1007: [BUG]  Rename and Organize
## Bug Description
I've caught a renaming issue for audio books. Mainly ordering issue for the naming. This happens when I select the  *Rename and Organize* option for audio books in downloads section of the settings.

## Steps To Reproduce

For a downloaded folder structure as the following, order is not respected in the processed result :

```
/Author/Title/
├── 00_Introduction
│   ├── 001_About.mp3
│   ├── 002_OtherBooks.mp3
│   ├── 003_Dedication.mp3
│   └── 004_FamilyTree.mp3
├── 01_Side 1
│   ├── 001.mp3
│   ├── 002.mp3
│   ├── 003.mp3
│   ├── 004.mp3
│   ├── 005.mp3
│   ├── 006.mp3
├── 02_Side 2
│   ├── 001.mp3
│   ├── 002.mp3
│   ├── 003.mp3
│   ├── 004.mp3
│   ├── 005.mp3
│   ├── 006.mp3
├── 03_Side 3
│   ├── 001.mp3
│   ├── 002.mp3
│   ├── 003.mp3
│   ├── 004.mp3
│   ├── 005.mp3
│   ├── 006.mp3
│   ├── 007.mp3
│   ├── 008.mp3
│   ├── 009.mp3
│   ├── 010.mp3
│   ├── 011.mp3
├── 04_Side 4
│   ├── 001.mp3
│   ├── 002.mp3
│   ├── 003.mp3
│   ├── 004.mp3
├── 05_Side 5
│   ├── 001.mp3
│   ├── 002.mp3
│   ├── 003.mp3
├── 06_Side 6
│   ├── 001.mp3
│   ├── 002.mp3
│   ├── 003.mp3
│   ├── 004.mp3
│   ├── 005.mp3
│   ├── 006.mp3
├── 07_Side 7
│   ├── 001.mp3
│   ├── 002.mp3
│   ├── 003.mp3
│   ├── 004.mp3
│   ├── 005.mp3
```

We have the correct total number of files, but for some reason the  `{partnumber}` suggested is not following the correct and expected order as in the structure above. My used template for renaming is like `{Author}/{Title}/{Title} {PartNumber}`.

I think some sorting is missing somewhere in the process, so that correct numbering could be applied...
I end up having my first file being `06_side/001.mp3`  instead of `00_introduction/001_About.mp3`.

mvanhorn:
## Summary

Audiobooks whose files are split across nested subfolders now keep their on-disk folder order when "Rename and Organize" assigns part numbers. Previously `natural_sort_key` keyed on `Path(path).name`, so identically named files from different folders (`001.mp3` in `00_Introduction/` vs `06_Side 6/`) collapsed together and the part sequence scrambled.

## Background

zazizou reported in #1007 that the total file count was right but `06_Side 6/001.mp3` came out as part 1 instead of `00_Introduction/001_About.mp3`. `scan.py` flattens every nested file into one list and `assign_part_numbers` sorts it with `natural_sort_key`, which discarded the parent-folder prefix. The fix pads the existing `PAD_NUMBERS_PATTERN` numbers over the full relative path (`str(path).lower()`) rather than the basename, so nested audiobooks sort by folder first and then by filename. Flat directories are unaffected because their files share a prefix, and `natural_sort_key` is only called from `assign_part_numbers`, so nothing else changes. I updated the test that pinned the old basename-only ordering and added nested-subfolder and duplicate-basename cases; `pytest tests/core/test_naming.py tests/core/test_part_number_extraction.py` passes (102 tests).

Closes #1007

### DIFF
--- a/shelfmark/core/naming.py
+++ b/shelfmark/core/naming.py
@@ -91,8 +91,7 @@ PAD_NUMBERS_PATTERN = re.compile(r"\d+")
 
 def natural_sort_key(path: str | Path) -> str:
     """Generate a sort key with padded numbers for natural sorting."""
-    filename = Path(path).name.lower()
-    return PAD_NUMBERS_PATTERN.sub(lambda m: m.group().zfill(9), filename)
+    return PAD_NUMBERS_PATTERN.sub(lambda m: m.group().zfill(9), str(path).lower())
 
 
 def assign_part_numbers(
--- a/tests/core/test_part_number_extraction.py
+++ b/tests/core/test_part_number_extraction.py
@@ -43,10 +43,10 @@ class TestNaturalSortKey:
             "file10.mp3",
         ]
 
-    def test_uses_filename_only(self):
+    def test_uses_full_path(self):
         files = [Path("/z/dir/file1.mp3"), Path("/a/dir/file2.mp3")]
         sorted_files = sorted(files, key=natural_sort_key)
-        assert sorted_files[0].name == "file1.mp3"
+        assert sorted_files == [Path("/a/dir/file2.mp3"), Path("/z/dir/file1.mp3")]
 
 
 class TestAssignPartNumbers:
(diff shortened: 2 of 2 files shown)


calibrain:
Nice ! Thank you for the contribution !

mvanhorn:
Thank you for the enthusiastic merge, @calibrain - audiobook part numbers follow subfolder order now.

PR #1136, opened 2026-07-24
Add Audiobook support to IRC

calibrain:


gmsiegel:
I updated to this build and tried a couple of audiobook searches, but don't get any results. Has anyone successfully found audiobooks with this new build? I also tried the dev build when that was recently released.

See my more detailed comments/info here:
https://github.com/calibrain/shelfmark/issues/1129

PR #1140, opened 2026-07-27
Keep Prowlarr results from distinct indexer entries separate

### LINKED ISSUE #1137: Prowlarr results deduplicated by `guid` alone, hiding results from filtered indexer entries on the same tracker
# Prowlarr results deduplicated by `guid` alone, hiding results from filtered indexer entries on the same tracker

**Version:** v1.3.4 (build `2026-07-21-1747b3466e730beff0b29019520bc636c3e7d36e`)

## Summary

When one tracker is configured in Prowlarr as several indexer entries that differ by search filter, Shelfmark collapses their results into a single release keyed on `guid`. Results from the filtered entries are discarded, so the user cannot see which releases matched the filter, and cannot choose which indexer entry a grab routes through.

The practical effect: with a "freeleech only" indexer entry alongside an unfiltered one, freeleech releases are invisible. They are silently replaced by the unfiltered entry's copy of the same torrent.

**This is not about freeleech wedges.** It applies to any tracker configured as multiple Prowlarr entries differing by server-side search filter, and to any filter type. Freeleech is simply the clearest example to demonstrate. Neither indexer entry used in the evidence below has wedge usage enabled.

## Where

`shelfmark/release_sources/prowlarr/source.py`, lines 890-899:

```python
for r in raw_results:
    key = (
        r.get("guid")
        or r.get("downloadUrl")
        or r.get("magnetUrl")
        or r.get("infoUrl")
        or f"{r.get('indexerId')}:{r.get('title')}"
    )
    if key in seen_keys:
        continue
    seen_keys.add(key)
    all_results.append(r)
```

`guid` is evaluated first, so `indexerId` and `downloadUrl` are never consulted when a `guid` is present, which is the normal case for Torznab results.

## Why this is wrong

`guid` identifies the **torrent**. It does not identify **which indexer entry surfaced it**, and on trackers that expose server-side search filters those entries carry real meaning.

MyAnonamouse exposes a `searchType` selector, so a common setup is several Prowlarr entries for one tracker:

| Prowlarr id | Name | searchType |
| --- | --- | --- |
| 10 | MyAnonamouse | Only active |
| 25 | MyAnonamouse - Freeleech Search | Freeleech |

Entry 25 returns only torrents that are currently freeleech. That is the whole reason it exists: it tells the user which releases cost nothing against ratio. Shelfmark discards that signal.

## Evidence

Querying Prowlarr directly for `Dune`:

```
idx 25 (Freeleech Search): 14 results
idx 10 (Only active):      20 results
shared guids:               5
```

Those 5 guids appear in both. In Shelfmark's release list only one row appears for each, sourced from whichever entry was queried first. There is no way to tell that those 5 are freeleech, and no way to grab via entry 25.

The same collapse applies to any pair of entries. On a broader query across four MAM entries, 20 guids were returned by more than one entry.

## Impact

- Freeleech (and other server-side filtered) releases are invisible in the UI
- Which copy survives depends on query ordering rather than user intent, so behaviour appears arbitrary
- The release row reports a single indexer, giving no indication alternatives were dropped
- Not MyAnonamouse specific. Any tracker configured as multiple Prowlarr entries has this shape

## Suggested fix

Include the indexer in the dedup key so results from distinct entries stay distinct:

```python
key = (
    r.get("indexerId"),
    r.get("guid")
    or r.get("downloadUrl")
    or r.get("magnetUrl")
    or r.get("infoUrl")
    or r.get("title"),
)
```

That alone restores correctness. Two optional improvements:

1. Show the indexer name on each release row, so duplicates are distinguishable rather than merely present
2. Allow an indexer preference order, so users who prefer one row per torrent get the entry they intended rather than the first queried

## Reproducing

1. Add the same tracker to Prowlarr twice, differing only in server-side search filter (for MAM: one `Only active`, one `Freeleech`)
2. Enable both in `PROWLARR_INDEXERS`
3. Search for a title present in both
4. Observe one row per torrent, always attributed to the same indexer, with no way to identify or select the filtered result

## Note

This is deliberately scoped to result visibility and indexer routing. MyAnonamouse freeleech *wedge* behaviour is a separate upstream matter (MAM changed the wedge mechanism; see Prowlarr#2673 and Prowlarr#2536) and is not what this issue is about.

delize:
Results were deduplicated on `guid` alone. When one tracker is configured in Prowlarr as several indexer entries differing only by a server-side search filter, all of them return the same guid for the same torrent, so every entry but the first was silently discarded. Freeleech and other filter-specific releases became invisible, replaced by the unfiltered entry's copy, and which copy survived depended on query ordering rather than user intent.

Include the indexer id in the dedup key so the entries stay distinct.

Release.source_id is qualified the same way. It keys the release cache and becomes the download task id, so rows sharing a guid would otherwise collide and a grab would route through whichever entry cached last, defeating the point of showing them separately. The handler's task matcher still accepts a bare guid or infoUrl so tasks queued before this change still resolve.

Ordering now follows the priority already configured in Prowlarr (1-50, lower preferred) rather than a new setting, since users curate that ranking there and filtered entries are typically ranked ahead of their unfiltered counterparts. The enabled-indexer list is fetched once and reused for the enrichment check, so this costs no extra round trip. Releases carry extra.indexer_priority, and the sort dropdown gains an "Indexer priority" entry, ascending, alongside the existing alphabetical "Indexer" sort; SortOption grew a default_direction for that, defaulting to desc so "Peers" is unchanged.

PROWLARR_COLLAPSE_DUPLICATES (default off) optionally collapses a release back to one row, resolved by the same Prowlarr priority. Left off, every entry that carried a release keeps its own row, which is what makes filtered results visible again.

Identity handling is defensive about partial payloads: a result that cannot be identified is never dropped or merged, and collapse only merges on a strong identifier (guid/downloadUrl/magnetUrl/infoUrl) because merging on title alone would discard genuinely different releases that share a name.

Fixes #1137

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -1211,6 +1211,7 @@ How long to cache individual book details. Default: 600 (10 minutes). Max: 60480
 | `PROWLARR_API_KEY` | Found in Prowlarr: Settings > General > API Key | string (secret) | _none_ |
 | `PROWLARR_INDEXERS` | Select which indexers to search. 📚 = has book categories. Leave empty to search all. | string (comma-separated) | _empty list_ |
 | `PROWLARR_AUTO_EXPAND` | Automatically retry search without category filtering if no results are found | boolean | `false` |
+| `PROWLARR_COLLAPSE_DUPLICATES` | Collapse a release that several indexer entries returned down to a single row. Leave this off to see every entry that carried it, which is what makes results from filter-specific entries (freeleech and the like) visible. | boolean | `false` |
 | `PROWLARR_USE_SEED_PREFERENCES` | Apply per-indexer seed time and ratio preferences from Prowlarr when sending torrents to the download client | boolean | `false` |
 
 <details>
@@ -1259,6 +1260,15 @@ Select which indexers to search. 📚 = has book categories. Leave empty to sear
 **Auto-expand search on no results**
 
 Automatically retry search without category filtering if no results are found
+
+- **Type:** boolean
+- **Default:** `false`
+
+#### `PROWLARR_COLLAPSE_DUPLICATES`
+
+**Show one row per release**
+
+Collapse a release that several indexer entries returned down to a single row. Leave this off to see every entry that carried it, which is what makes results from filter-specific entries (freeleech and the like) visible.
 
 - **Type:** boolean
 - **Default:** `false`
--- a/shelfmark/release_sources/__init__.py
+++ b/shelfmark/release_sources/__init__.py
@@ -163,6 +163,7 @@ class SortOption:
 
     label: str  # Display label in the sort dropdown
     sort_key: str  # Field to sort by on the Release object
+    default_direction: Literal["asc", "desc"] = "desc"  # Which way "best first" runs
 
 
 @dataclass
@@ -261,7 +262,12 @@ def serialize_column_config(config: ReleaseColumnConfig) -> dict[str, Any]:
     # Include extra sort options (sort entries not tied to a column)
     if config.extra_sort_options:
         result["extra_sort_options"] = [
-            {"label": opt.label, "sort_key": opt.sort_key} for opt in config.extra_sort_options
+            {
+                "label": opt.label,
+                "sort_key": opt.sort_key,
+                "default_direction": opt.default_direction,
+            }
+            for opt in config.extra_sort_options
         ]
 
     # Include action button if specified (replaces default expand search)
--- a/shelfmark/release_sources/prowlarr/api.py
+++ b/shelfmark/release_sources/prowlarr/api.py
@@ -189,16 +189,24 @@ class ProwlarrClient:
         indexers = self.get_indexers(raise_on_error=raise_on_error)
         return [idx for idx in indexers if idx.get("enable", False)]
 
-    def get_enriched_indexer_ids(self, *, restrict_to: list[int] | None = None) -> list[int]:
+    def get_enriched_indexer_ids(
+        self,
+        *,
+        restrict_to: list[int] | None = None,
+        indexers: list[dict[str, Any]] | None = None,
+    ) -> list[int]:
         """Return enabled indexer IDs that benefit from extra Torznab handling.
 
         Args:
             restrict_to: Optional list of candidate indexer IDs to consider.
+            indexers: Optional already-fetched enabled indexer list, so callers
+                that need the full records for other reasons can avoid a second
+                round trip.
 
         """
         enriched_ids: list[int] = []
 
-        for idx in self.get_enabled_indexers_detailed():
+        for idx in indexers if indexers is not None else self.get_enabled_indexers_detailed():
             idx_id_int = coerce_int_like(idx.get("id"))
             if idx_id_int is None:
                 continue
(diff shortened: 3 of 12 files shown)


delize:
## How MyAnonamouse is structured across Prowlarr entries

Some context for reviewers on the shape of the problem, and on why the `source_id` change is part of this PR rather than a separate concern.

### 1. The structure

MyAnonamouse is a **single tracker**. Prowlarr lets you register it several times, each entry differing only by a server-side search filter. Each registration gets its own **indexer ID** and its own **priority** (1-50, lower is preferred).

| Indexer ID | Entry name | Server-side filter | Priority |
|---|---|---|---|
| 25 | MyAnonamouse - Freeleech Search | `searchType=freeleech` | **20** |
| 8 | MyAnonamouse - Freeleech Token | `searchType=fl-token` | **21** |
| 9 | MyAnonamouse - VIP & Freeleech Token | `searchType=vip-fl` | **21** |
| 10 | MyAnonamouse | `searchType=active` | 24 |

Note the priorities: the filtered entries are already ranked ahead of the unfiltered one. That ranking is why this PR reuses Prowlarr's own priority rather than adding a second one in Shelfmark.

### 2. What one search produces

Each entry is queried separately. The same physical torrent comes back from more than one of them, carrying **the same `guid`** every time, because the guid identifies the torrent, not the entry that surfaced it.

```mermaid
flowchart LR
  MAM[("MyAnonamouse<br/>one tracker")]
  MAM --> I25["idx 25 · prio 20<br/>Freeleech Search"]
  MAM --> I8["idx 8 · prio 21<br/>Freeleech Token"]
  MAM --> I10["idx 10 · prio 24<br/>Only active"]
  I25 --> G1["guid: tor/1001<br/>dvf=0 · FREELEECH"]
  I8  --> G2["guid: tor/1001<br/>token eligible"]
  I10 --> G3["guid: tor/1001<br/>dvf=1 · standard"]
  G1 --> SAME{{"same guid<br/>three entries"}}
  G2 --> SAME
  G3 --> SAME
```

### 3. Before and after

**Before**, keyed on `guid` alone: one row survives, whichever entry was queried first. The freeleech and token rows are discarded silently, taking the freeleech signal with them.

**After**, keyed on `(indexerId, guid)`: three keys, three rows, each keeping its own flags and its own grab route.

`source_id` had to change alongside it. It keys the release cache and becomes the download task id, so three rows sharing one guid would collide there and every grab would route through whichever entry cached last. That would have defeated the second half of the issue (choosing which entry a grab goes through), which is why it is in this PR rather than a follow-up.

### 4. What decides the order you see

```mermaid
flowchart TD
  R["results from all entries"] --> P["backend: sort by Prowlarr priority<br/>20 → 21 → 24"]
  P --> E["backend: enriched indexers first<br/>within equal priority"]
  E --> UI{"sort selected in the UI?"}
  UI -->|"Default / best match"| REL["frontend: title + author relevance<br/>PRIMARY key"]
  REL --> TIE["ties broken by backend order<br/>= priority · SECONDARY key"]
  UI -->|"Indexer"| ALPHA["alphabetical by name"]
  UI -->|"Indexer priority"| PRIO["Prowlarr priority, lowest first"]
```

Relevance stays the primary key on purpose: a better-priority indexer should not float the wrong edition above the book that was actually searched for. A search returning two editions therefore shows freeleech-first *twice*, once inside each edition's group, rather than bunching all freeleech rows at the top.

The existing "Indexer" sort is alphabetical by name, which puts `MyAnonamouse` above `MyAnonamouse - Freeleech Search`. That is unchanged; the new "Indexer priority" entry is what sorts on the priority directly.

calibrain:
Awesome ! Thank you again !

PR #1142, opened 2026-07-27
Add a {Language} naming template variable, and consolidate language resolution

### LINKED ISSUE #1138: [Feature Request] `{Language}` naming template variable, so different-language editions of the same book do not collide
# [Feature Request] `{Language}` naming template variable, so different-language editions of the same book do not collide

**Version:** v1.3.4 (build `2026-07-21-1747b3466e730beff0b29019520bc636c3e7d36e`)

## Problem

Grabbing two language editions of the same book puts both files in one folder, because the naming template can only reference the metadata provider's canonical title. Both editions resolve to the same book, so both render to the same path.

Real example. English and Swedish editions of *Project Hail Mary*, both grabbed from MyAnonamouse:

```
Andy Weir/Project Hail Mary/Andy Weir - Project Hail Mary.m4b      16:10:53   924.6 MB   (English)
Andy Weir/Project Hail Mary/Andy Weir - Project Hail Mary_1.m4b    20:15:30   589.8 MB   (Swedish)
```

Template used: `{Author}/{Title}/{Author} - {Title}{ (PartNumber)}`

The `_1` suffix is collision avoidance, so no data is lost. But the two files are now one library item downstream. Audiobookshelf reads that folder as a single book with two audio tracks and a 36 hour runtime, and the only fix is to separate them by hand.

The files themselves are correctly tagged and perfectly distinguishable:

```
Andy Weir - Project Hail Mary.m4b     TAG:album = Project Hail Mary
Andy Weir - Project Hail Mary_1.m4b   TAG:album = Uppdrag Hail Mary – Ensam i rymden
```

## Shelfmark already knows the language

This is the frustrating part. The language is parsed, stored and displayed, it just never reaches the template engine.

`shelfmark/release_sources/prowlarr/source.py` extracts it from MAM titles like `[ENG / M4B]` and `[SWE / M4B]`, maps it to a 2-char ISO code, and attaches it to the release:

```python
MAM_LANGUAGE_MAP = {"eng": "en", "ita": "it", "spa": "es", "fra": "fr", ...}

def _extract_mam_language(raw_title: str) -> str | None: ...

language_detected = _extract_mam_language(str(raw_title or ""))
Release(..., language=language_detected, ...)
```

The release list renders this as the EN / SV badges next to each result, so the UI is already language-aware at selection time. The information is simply dropped before naming.

## Requested change

Expose the release language as a `{Language}` template variable, so a template such as:

```
{Author}/{Title}{ (Language)}/{Author} - {Title}{ (PartNumber)}
```

produces:

```
Andy Weir/Project Hail Mary/...
Andy Weir/Project Hail Mary (sv)/...
```

Using the arbitrary prefix/suffix syntax from #559, `{ (Language)}` emits nothing when the language is unknown, so existing layouts are unaffected and users who do not want it simply leave the token out.

## Implementation notes

Four touchpoints, following the pattern already used for series and user data:

1. `shelfmark/core/models.py` — add `language: str | None = None` to `DownloadTask`. The neighbouring fields are already annotated for exactly this purpose (`# Series info (for library naming templates)`, and `username` is commented `# Username for {User} template variable`)
2. Populate it when the task is created from a `Release`
3. `shelfmark/download/postprocess/transfer.py` — add `"Language": task.language` to `build_metadata_dict()`
4. `shelfmark/core/naming.py` — add `"language"` to `KNOWN_TOKENS`

Plus a `{Language}` chip in the frontend "Insert variable" picker.

## Caveat worth stating up front

Language extraction is currently MyAnonamouse specific (`_extract_mam_language`). For AudiobookBay, Anna's Archive and other sources the value would be empty, and `{ (Language)}` would render as nothing. That still solves the collision for the source where it actually happens, and other sources can populate `Release.language` later without further template work.

## Alternatives considered

- **`{OriginalName}` in the folder segment.** Does separate the editions, since the release filenames differ, but names every folder after whatever the uploader called the file, which is considerably worse for the common single-edition case.
- **Restricting downloads to one language.** Defeats the point of deliberately keeping more than one.
- **Manual separation after the fact.** What I do today. Workable, but it needs noticing, and a silent `_1` suffix is easy to miss until the runtime looks wrong in the library.

### LINKED ISSUE #1141: Consolidate the three language-alias maps onto the bundled language database
## Problem

Three release sources each maintain their own language-alias map, all resolving to the same ISO 639-1 codes, and all sitting alongside a shared 70-entry database that already exists in the repo but that only one of them uses.

| Where | Aliases | Codes | Source of truth |
|---|---|---|---|
| `data/book-languages.json` | — | **70** | the bundled DB, `{"language": "English", "code": "en"}` |
| `shelfmark/release_sources/direct_download.py` | loads the DB lazily via `_language_alias_to_code()` | 70 | uses the DB |
| `shelfmark/release_sources/prowlarr/source.py` `MAM_LANGUAGE_MAP` | 41 | 33 | hardcoded dict |
| `shelfmark/release_sources/audiobookbay/source.py` `LANGUAGE_MAP` | 33 | 33 | hardcoded dict |

The two hardcoded maps have to be edited in lockstep with each other and kept consistent with the DB by hand.

## Current state is not broken

Worth stating up front, since this is hygiene rather than a bug. Cross-checking the three today:

- Every code Prowlarr emits is present in `data/book-languages.json`. No orphans.
- Every code AudiobookBay emits is present too. No orphans.
- Prowlarr correctly handles the ISO 639-2/B vs 639-2/T split (`ger`/`deu`, `fre`/`fra`, `cze`/`ces`, `dut`/`nld`, `gre`/`ell`, `rom`/`ron`), which is the part most likely to be got wrong.

So nothing is currently mismapped. The risk is drift the next time someone adds a language to one map and not the others.

## Suggestion

Promote the alias-resolution logic out of `direct_download.py` into something shared, e.g. `shelfmark/core/languages.py`:

- `normalize_language(value) -> str | None`, accepting a 2-letter code, a 3-letter code (both /B and /T), or a full English name, and returning the canonical ISO 639-1 code.
- Backed by `data/book-languages.json`, with the 3-letter aliases currently living in `MAM_LANGUAGE_MAP` folded into that file so there is one place to add a language.
- `direct_download.py`, `prowlarr/source.py` and `audiobookbay/source.py` then all call it, and the two hardcoded dicts go away.

Deliberately not part of #1138 (the `{Language}` naming token): that change only needs `Release.language` to already hold a code, which it does. This is a separate cleanup and touches hot, well-tested source files, so it deserves its own review.

## Notes for whoever picks it up

- `direct_download.py` guards the DB load with a lock and a module-level cache, and has an `_AMBIGUOUS_SHORT_LANGUAGE_CODES` set (`de`, `en`, `it`, `la`, `no`, `or`, `is`, `in`) that only accepts those codes in bracket/key context, because they collide with common English words when parsing free text. That heuristic is specific to path parsing and probably should **not** move into a general normalizer — worth keeping the parsing and the normalizing separate.
- 37 of the DB's 70 codes cannot currently be produced by any source (`af`, `az`, `bn`, `fa`, `ga`, `gl`, `hy`, `jv`, …). Consolidating does not by itself make those reachable; that needs the sources to emit them.
- `src/frontend/src/data/languages.ts` carries a 10-entry `DEFAULT_LANGUAGES` fallback whose comment says "This data is loaded from the backend in production". Worth checking whether that fallback is still needed once there is one canonical backend list.

delize:
Fixes #1138
Fixes #1141

## Problem

Two language editions of one book resolve to the same canonical title, so they render to the same path and the second gets a `_1` collision suffix. Audiobookshelf treats a folder as exactly one library item, so the pair becomes a single book with both files as tracks and a summed runtime.

Shelfmark already parses and displays the language. It just never reached the template engine.

## `{Language}` template variable

A template like `{Author}/{Title}{ (Language)}/{Author} - {Title}` now yields:

```
/library/J K Rowling/Harry Potter (sv)/J K Rowling - Harry Potter.m4b
/library/J K Rowling/Harry Potter/J K Rowling - Harry Potter.m4b
```

The untagged edition's path is byte-identical to today, so no existing layout shifts.

Three details worth flagging:

**The value is casefolded.** On a case-insensitive filesystem `(SV)` and `(sv)` would collapse back into one folder, reintroducing the exact collision being fixed.

**Values meaning "we don't know" render nothing** rather than producing `Project Hail Mary (unknown)` folders. Anna's Archive reports that string literally (`direct_download.py`, `language = detected or "unknown"`).

**The frontend wasn't sending the release language at all**, so the token would have stayed empty for exactly the audiobook sources in the report. Prowlarr and AudiobookBay do not put language in `extra` the way `direct_download` does, hence the payload plumbing. It reads `release.language`, never `book.language` — the latter is the provider's canonical edition and would mislabel a translation, with a regression test for that specifically.

Not gated to audiobooks: Calibre-Web-Automated stages ingested files by basename and discards folder structure, so the rename (filename) template is the only lever those users have. Verified that form works: `J K Rowling - Harry Potter (sv).epub`.

## Language consolidation (#1141)

Three release sources each carried their own alias map, all resolving to the same ISO 639-1 codes, alongside a bundled database that only one of them used. Adding a language meant editing three places.

Aliases now live in `data/book-languages.json` beside the code and name they belong to, and `shelfmark/core/languages.py` resolves any of them — two-letter code, ISO 639-2 three-letter in either the bibliographic or terminological form, or English name. Prowlarr and AudiobookBay drop their tables. Direct Download keeps its own path-parsing heuristics, including the ambiguous short codes that collide with English words (`de`, `en`, `no`, `in`), and takes only the alias data.

This also closes a coverage gap. MyAnonamouse offers 62 languages; Prowlarr mapped 37, and an unmapped code is *dropped* rather than passed through, so the other 25 carried no language at all — leaving `{Language}` empty and the collision unfixed for Latin, Farsi, Tamil, Urdu and the rest. Seven languages MAM offers had no database entry at all: Bosnian, Burmese, Estonian, Icelandic, Manx, Scottish Gaelic, Sanskrit.

Also fixes the Traditional Chinese code, which used a U+2011 non-breaking hyphen. Nothing compares against the ASCII spelling today so it was latent, but it would silently defeat the first thing that did.

## Validation

Verified end to end against a live Prowlarr and MyAnonamouse, not just unit tests. A real search returning both an English and a Swedish edition, through the actual `queue_release` → `DownloadTask` → naming path:

```
STEP 1  real MAM search        -> 37 releases, languages: ['en', 'sv']
STEP 3  queue_release          -> task.language='sv'
STEP 4  build_metadata_dict    -> metadata['Language']='sv'
STEP 5  build_library_path     -> /library/J K Rowling/Harry Potter (sv)/...
two language editions resolve to DIFFERENT folders: True
```

The refactor is pinned by a snapshot of both per-source maps taken *before* they were deleted. All 131 aliases are asserted to still resolve to the same code, one parametrised test each, so a regression names the specific alias.

Also verified: the filename-only template, the retry round-trip (`serialize_task_for_retry` → `_restore_task_from_retry_payload`, plus a legacy payload with no `language` key), and placeholder handling.

Added a `KNOWN_TOKENS` ordering invariant test — `find_placeholder()` does a substring `.find()` in list order and nothing protected that contract, so a future token in the wrong position could silently shadow an existing one. And a lockstep guard on the frontend, since `KNOWN_TOKENS` is hand-duplicated in TypeScript.

**One caveat worth stating.** Three MAM codes are confirmed by observation (`ENG`→`en`, `SWE`→`sv`, `MAL`→`ml`, the last from a real `[MAL / EPUB]` Tagore release). The remaining ~59 are derived from ISO 639-2 rather than observed, because MAM's catalogue is overwhelmingly English — enabling 27 extra languages still yielded only one non-English hit across 258 results. Mitigated rather than closed: both 639-2 variants are present for every language where they differ, and a wrong alias is an unused entry while a missing one loses the language. Happy to correct any code a maintainer knows differs.

## Test results

2056 Python tests pass (up from 1906). Frontend typecheck, lint, format and 126 unit tests pass.

Pre-existing failures on my machine, unchanged by this branch and unrelated: `tests/bypass/` needs `seleniumbase`, and `tests/config/test_entrypoint_permissions.py` uses bash-4 syntax that macOS bash 3.2 rejects.


delize:
Follow-up after auditing everything else that touches the language data, since consolidating it changed a shared file's schema.

**Found and fixed one leak.** `settings.py` read `data/book-languages.json` directly and `main.py` served the parsed result straight to the frontend as `book_languages`, so adding aliases beside each entry leaked them into every `/api/config` response — 42% larger (3044 → 4319 bytes), carrying a field the frontend's `Language` type does not declare. Both now go through `supported_book_languages()`, which returns only `{code, language}`, with a test pinning that shape so a future resolution field cannot leak the same way.

**Checked and fine:**

- **Default Book Languages setting** — options are derived from the same list, so the seven added languages become selectable automatically. 77 options, no aliases in them.
- **Release modal language filter** — `buildLanguageNormalizer` is driven by the API list, so it picks up new languages with no change.
- **Language badge colours** — `getLanguageColor` falls back to a default for unmapped codes. The seven new languages have no dedicated colour, but neither do ~49 of the languages that were already there, so this is pre-existing and cosmetic rather than a regression.
- **No other consumer** of `book-languages.json` exists; those two were the only ones.

calibrain (changes requested):
Hey, thank you so much, this is great !
From a quick review, it seems there might be a bug where if for a book we have (en) or (english) we would have duplicate folders
Can we do something about that ? Just forcing a normalization maybe ?


Another thing, for Traditional Chinese, we have `zh‑Hant` but in the new PR we get `zh-Hant` instead, (note the `-` is not the same) 
This would break current customers who are using that language. Could you add an alias or reuse the older `-` ? 

Thank you again !

delize:
Hey, updated for both occurrences. Let me know what you think. 

Also added uniform on Unicode characters and dashes so hopefully it doesn't occur again for other situations. 

Also send I might have over engineered, on GitHub mobile, that shows as an underscore in a code block 😅 not as a dash.

calibrain:
This is great work, thank you for the contribution !

Issue #1035, opened 2026-05-30
[FEATURE REQUEST] Plugin for custom acquisition sources


Hamartia-Ate:
Would love a plugin script of some kind that allows external scripts or services to act as acquisition sources. This lets users hook in scripts for niche sources that would never warrant a built-in integration, without anything needing to be merged upstream.

Thanks

Issue #595, opened 2026-02-07
[Feature] Search via Indexer instead of Metadata


Vallaquenta:
I was wondering if it's possible to change the search to not use any metadata, but immediately query an indexer (like Prowlarr) and display results from that.

raygan:
I'd like this option as well, at least as a fallback we can manually enable. For instance here I'm searching for a book that doesn't yet exist in Hardcover or Open Library. The only results are other similarly named but incorrect books:

<img width="2267" height="1232" alt="Image" src="https://github.com/user-attachments/assets/7e0f4ded-304d-4109-a2c1-bf5e6a2dd070" />

If I want to download the book, I have to select an incorrect book, then click the pencil icon to edit the search query, then choose the download.

<img width="2133" height="1139" alt="Image" src="https://github.com/user-attachments/assets/9615a8b3-a841-45ba-a63a-d574d0dc36c6" />

Ideally I'd like to see:

- A "Search Directly" button on the search results page, to use as a fallback when metadata search is poor, and
- A "Search Directly (No Metadata)" option in the Search Mode settings, to search this way by default.

PR #1172, opened 2026-08-08
feat(metadata): add Moly.hu metadata provider

FlyinPancake:
First of all, I don't know if you even want to merge a scraper-based metadata provider. I made this just for my use-case. If you'd rather not, I completely understand it.

An alternative would be adopting [Audiobookshelf's Metadata Provider API](https://audiobookshelf.org/docs/documentation/community/community-providers) which I contributed to it for exactly the reason to not have scrapers. 

## What

Adds [Moly.hu](https://moly.hu) — the Hungarian community book catalog — as a metadata provider, following the existing provider plugin architecture (`@register_provider` + settings tab with enable checkbox and Test Connection button, disabled by default).

## Why

None of the current providers cover Hungarian editions well: Hardcover and Open Library rarely index them, and Google Books coverage is spotty. Moly.hu is the de-facto catalog for Hungarian books (local editions *and* Hungarian translations of foreign works). With this provider, Universal mode works end-to-end for Hungarian titles: moly search → localized title/author feed the release search → indexers that carry Hungarian content can actually match. Related pain points: #595 (books missing from metadata providers), #1035 (interest in niche sources).

## How

- HTML scraping with BeautifulSoup (already a dependency), no API key needed
- Scraping approach (search URL, page structure, language-tag mapping) adapted from the long-lived Calibre `Moly_hu` plugin (GPL v3, credited in the module docstring), with fallback selector chains inherited from it
- Sliding-window rate limit (30 req/min) to stay polite to a small community site
- Standard `@cacheable` decorators; fetch failures return `None` so they are not cached (same behavior as the Google Books provider)
- Search results carry cover thumbnails, rating and series info as display fields; `get_book` parses title (zero-width chars stripped, nested series link excluded), authors, ISBN-13/10, publisher, publish year, description (spoiler-warning prefix stripped), tags/genres, cover, and language (from moly's language tags, defaulting to `hu`)
- ISBN search resolves through moly's site search

## Testing

- `tests/metadata/test_moly_parse.py`: offline tests with fixture HTML mirroring live moly.hu markup — search parsing/dedup, pagination guard, failure-not-cached behavior, book-page parsing, ISBN resolution, ISBN validation helper
- `uv run pytest tests/metadata` green (45 passed), `ruff check` / `ruff format` clean
- Verified live against moly.hu (search, get_book, ISBN lookup) and running in Docker alongside Hardcover

### DIFF
--- a/shelfmark/metadata_providers/__init__.py
+++ b/shelfmark/metadata_providers/__init__.py
@@ -709,3 +709,6 @@ with suppress(ImportError):
 
 with suppress(ImportError):
     from shelfmark.metadata_providers import googlebooks as googlebooks
+
+with suppress(ImportError):
+    from shelfmark.metadata_providers import moly as moly
--- a/shelfmark/metadata_providers/moly.py
+++ b/shelfmark/metadata_providers/moly.py
@@ -206,13 +206,14 @@ class MolyProvider(MetadataProvider):
 
         fields_key = ":".join(f"{k}={v}" for k, v in sorted(options.fields.items()))
         cache_key = f"{query}:{options.search_type.value}:{options.limit}:{fields_key}"
-        return self._search_cached(cache_key, query, options.limit)
+        return self._search_cached(cache_key, query, options.limit) or []
 
     @cacheable(ttl_key="METADATA_CACHE_SEARCH_TTL", ttl_default=300, key_prefix="moly:search")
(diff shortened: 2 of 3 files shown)


calibrain:
Ideally we would put this behind some flag that would make it available only if Hungarian language is selected, but since its an opt-in anyway, I am fine including this
Thank you very much for the contribution !

PR #1146, opened 2026-07-29
feat(download): add native Real-Debrid and AllDebrid debrid service clients

PaulVumon:
## Summary
Adds native download client support for debrid services (**Real-Debrid** and **AllDebrid**), routing magnet links and torrents through their APIs to download content via their CDN infrastructure.

## Features
- **Real-Debrid Client (\RealDebridClient\)**:
  - API v1.0 integration (magnet upload, file selection, status polling, link unrestricting, HTTP download)
  - Premium account validation via connection test
- **AllDebrid Client (\AllDebridClient\)**:
  - API v4/v4.1 integration (magnet upload, status polling, file tree listing, link unlocking with delayed link support, HTTP download)
  - Multi-tier link resolution strategy with automatic fallback
  - Premium account validation via connection test
- **Common & UI**:
  - Ebook & Audiobook file filtering (\.epub\, \.m4b\, \.mp3\, \.pdf\, \.flac\, etc.)
  - Settings UI options in \PROWLARR_TORRENT_CLIENT\ dropdown with API key configuration and test connection buttons

PR #1173, opened 2026-08-08
Feat: Add torbox download client

droodroo-blip:
## AI Use Notice
AI was used to co-author the changes, that were then reviewed by me. PR description was written solely by me.

## Context
Based on earlier PR #1146, here is PR to add support for using Torbox in two capacities:
1. As a torrent download client; this works almost identically to the real-debrid and alldebrid integrations from that earlier PR.
2. As a direct download enabler; this is a bit more tricky, and I recognize that it conflicts with the idea of a "direct download", but Torbox offers the feature to download contents from Anna's archive for you. This simplifies the download process and offers a faster solution that isn't reliant on any cloudflare bypass mechanism. I wanted to see if I could integrate this within another part of shelfmark, but this was the most obvious way to do it since we can leverage the existing web downloading logic around using AA.

I'm mostly creating this for my own use, but figured I would put in a PR if this is deemed to be useful for others. Thanks for creating a kick-ass project and thanks for taking the time to read this description!

### DIFF
--- a/shelfmark/config/settings.py
+++ b/shelfmark/config/settings.py
@@ -1498,6 +1498,15 @@ def download_source_settings() -> list[SettingsField]:
             label="Account Donator Key",
             description="Enables fast download access on AA. Get this from your donator account page.",
         ),
+        CheckboxField(
+            key="TORBOX_DIRECT_DOWNLOAD_ENABLED",
+            label="Use Torbox for Direct Downloads",
+            description=(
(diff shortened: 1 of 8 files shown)


droodroo-blip:
Converting to draft for now, I encountered some frontend errors while testing

calibrain:
Thank you for the contribution, let me know when its ready for review, ill try to carve some time to take a proper look !

PR #1203, opened 2026-08-14
fix: sync renamed CWA usernames safely

### LINKED ISSUE #1197: [BUG] CWA-linked users: username never syncs after being renamed in CWA, and UI blocks manual edit
**Version:** v1.3.5 (build 2026-07-31-cdd156e1427e32cb47f9ac7c3bfa25814c7c66f2)
**Auth mode:** AUTH_METHOD=cwa

**Bug**

When a user's username changes on the Calibre-Web-Automated (CWA) side, Shelfmark's CWA sync correctly detects the identity via email match (`unique_email_match`) but never updates the local `username` field. `_build_updates()` in `shelfmark/core/external_user_linking.py` only writes `auth_source`, `role`, `email`, `display_name`, and `oidc_subject` — `username` is never included, so it stays frozen at whatever value existed the first time the user was linked.

This is compounded by the Admin -> User Management UI, which disables the username field for externally-linked (`auth_source=cwa`) users — presumably because it's expected to stay synced from the source. Since it never actually does, there's no supported way to fix a stale username at all; the only workaround is editing `users.db` directly.

**Repro**

1. Set `AUTH_METHOD=cwa`, sync a CWA user into Shelfmark.
2. Rename that user in CWA.
3. Click "Sync CWA Users" in Shelfmark. Log shows the match reason flip from `existing_source_username_match` to `unique_email_match`, and "Synced N CWA users (0 created, N updated, 0 deleted)" — but the Shelfmark-side `username` is unchanged.
4. Try editing the username directly in Shelfmark's own Admin UI for that user — field is disabled because `auth_source=cwa`.

**Suggested fix**

Either include `username` in the fields `_build_updates()` writes on sync (with a collision-handling strategy, since two synced users could theoretically collide on rename), or — if immutability is intentional (e.g. for session/FK stability) — allow manual username edits in the Admin UI for CWA-linked users so there's at least a supported escape hatch.

**Workaround used**

Stopped the container and ran `UPDATE users SET username='newname' WHERE id=N;` directly against `users.db`.

floze-the-genius:
## Summary

- sync an existing CWA-backed user's username when CWA renames it
- keep username collisions safe by assigning a stable `__cwa` alias instead of overwriting a local account
- allow username updates through `UserDB` and cover rename/collision/repeat-sync behavior

Fixes #1197.

## Testing

- `uv run ruff check shelfmark tests`
- `uv run ruff format --check shelfmark tests`
- `uv run vulture shelfmark`
- `uv run pytest tests/core/test_cwa_user_sync.py tests/core/test_user_db.py tests/core/test_admin_users_api.py tests/core/test_auth_api.py -k "cwa or update_user"` (36 passed)
- `uv run pytest tests/ -x --tb=short -m "not integration and not e2e" --ignore=tests/config/test_entrypoint_permissions.py -q` (2445 passed, 5 skipped)

The entrypoint permission tests were excluded locally because macOS ships Bash 3.2, which does not support the `${1,,}` expansion used by `entrypoint.sh`; the same failure reproduces on an unchanged checkout. `make python-typecheck` also currently reports the existing `settings.py:147` callback return-type mismatch on the unchanged base.


calibrain:
Thank you for the contribution ! I really appreciate it

CalmStilton:
Lovely. Thank you

Issue #1220, opened 2026-08-15
[BUG] Direct Downloads failing intermittently


crispybegs:
Hiya, my direct downloads are failing most of the time but work occasionally. I fed my logs into an ai (boo hiss) to see what it had to say and it came back with this:

**I can see exactly what happened:**

**First search (The Stand):**
Got 403 → triggered native bypasser → solved DDoS-Guard → extracted 9 cookies → found 3 releases ✅

**Second search (Amrita, ~2 minutes later):**
Shelfmark tried to reuse the stored cookies from the first search
Those cookies were already stale/invalid
DDoS-Guard redirected to check=1 instead of serving the page
Shelfmark followed the redirect but didn't re-trigger the bypasser (it only triggers on 403, not on redirect loops)
Result: TooManyRedirects loop until it gives up ❌

**The problem**
This is a Shelfmark bug/limitation. The native bypasser successfully solves DDoS-Guard, but the extracted cookies have a very short TTL (likely single-use or time-bound to that specific request). When Shelfmark tries to reuse them for a subsequent search, they're expired, and the bypasser doesn't re-activate because it only watches for 403s — not redirect loops.

**What this means**
Your setup is actually correct and working. The VPN routes through Sweden. The bypasser can solve DDoS-Guard. But Shelfmark's cookie persistence is broken for Anna's Archive right now.

**What you can do**
**Option 1:** Report it to the Shelfmark developer
This is clearly a bug — the bypasser should either:
Re-trigger when it hits a redirect loop (not just 403), or
Not cache DDoS-Guard cookies across searches
Open an issue at github.com/calibrain/shelfmark with these logs. The developer can fix this.

**Option 2:** Retry searches until they work
Since the bypasser does work on the first attempt after a fresh start, you could restart Shelfmark between searches.
This forces a fresh bypasser session. Annoying, but functional.

**Option 3:** Use Prowlarr as primary
Your Prowlarr is configured and working. Disable Anna's Archive in Shelfmark's download sources, rely on Prowlarr for releases, and manually grab from Anna's Archive via browser when Prowlarr comes up empty.

**Option 4:** Wait for a Shelfmark update
The latest tag will auto-update when the developer fixes this.
I wish I had a compose-level fix for you, but this is inside Shelfmark's bypass logic. The infrastructure (VPN, DNS, mirrors, bypasser engine) is all correct — it's just the cookie reuse that's broken.

is it chatting breeze or is it correct about anything here?

[...]

PR #1221, opened 2026-08-15
fix(bypass): discard rejected DDoS-Guard cookies instead of replaying them

calibrain:
A cookie that has been rejected was kept and presented again on every later
request, so a single bad clearance could re-arm the challenge indefinitely.

Cookie storage:
- Enforce expiry for every stored cookie, not just cf_clearance. DDoS-Guard
  domains have no cf_clearance, so the existing check never fired for them and
  expired cookies were replayed forever.
- Stop storing the per-check cookies __ddg8_/__ddg9_/__ddg10_ and
  ddg_last_challenge. Captured live from Anna's Archive, these carry the client
  IP and the timestamp the check was issued (~40 min), versus ~1 year for the
  __ddg1_/__ddg2_/__ddgid_ clearance. Replaying an IP-bound token stops
  describing the caller as soon as the egress IP changes, which is routine
  behind a VPN.

Failure handling — every path that is rejected while carrying cookies now
purges them, not just the redirect loop:
- 403 returned while presenting cookies.
- Cached-cookie attempt rejected, whether by status or by redirect loop.
- Factored the purge into _purge_clearance, guarded on a non-empty hostname
  since clear_cf_cookies("") means "every host" and would wipe clearance for
  sites that are working fine.

Also fix the search warm-up switches shipped inert in v1.3.8:
SEARCH_WARMUP_ENABLED and SEARCH_WARMUP_QUERY are not in the settings
registry, and config.get only consults the environment for keys it knows, so
both always returned their defaults — the warm-up could not be turned off or
retargeted. Read os.environ first.

Refs #1220. Deliberately not "Fixes": the reported failure could not be
reproduced on v1.3.8 from a stable IP (the reporter's own queries all returned
200 on both the pre- and post-change builds), and the new purge paths did not
fire in live testing because the failures arrive as redirect loops, which were
already purged. These are correctness fixes with no measured effect on that
issue. The underlying problem remains that Chrome-obtained cookies never
satisfy DDoS-Guard when replayed by requests, so every search still re-solves.

Verified: 2542 unit tests pass; ruff, basedpyright and vulture clean; e2e
platform baseline (10), full (6) and bypasser-external (5) all pass; five
sequential live searches against Anna's Archive all returned 200 with zero
"Exceeded 30 redirects".

### DIFF
--- a/shelfmark/bypass/internal_bypasser.py
+++ b/shelfmark/bypass/internal_bypasser.py
@@ -253,6 +253,26 @@ DDG_COOKIE_NAMES = {
     "ddg_last_challenge",
 }
 
+# DDoS-Guard cookies that describe *one* check rather than granting clearance, and so
+# must never be replayed on a later request. Observed live on Anna's Archive:
+#
+#   __ddg9_   the client IP address
+#   __ddg10_  the unix timestamp the check was issued
+#   __ddg8_   an opaque token issued with them, same ~40 minute expiry
+#
+# Clearance itself lives in __ddg1_/__ddg2_/__ddgid_ (roughly a year) and __ddg5_.
+# Replaying the trio is actively harmful: once the timestamp ages out - or the egress
+# IP changes, which happens routinely behind a VPN - the values no longer describe the
+# caller, DDoS-Guard re-arms its check and answers every request with a ?check=1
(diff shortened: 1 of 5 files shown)


calibrain:
Fix #1220

Issue #1225, opened 2026-08-17
README still says the project is not under active maintenance


XeroIP:
Hi! Small documentation thing rather than a bug.

The readme still carries this note:

> [!NOTE]
> This project is in a stable state as of May 2026 but is not under active maintenance.

That looks like it was simply left behind rather than intended. It was added in #1000 on 2026-05-14, which was @alexhb1's last commit before stepping down, and you clarified the situation afterwards in [Discussion #1045](https://github.com/calibrain/shelfmark/discussions/1045):

> I am the original Dev, and for a while, @alexhb1 was the maintainer, and did a phenomenal job, but then had to step down. [...] I won't archive it, as I plan on still monitoring security alerts and merging PR's if anyone wants to help.

The activity since then goes well beyond that description. As of today:

| Signal | Value |
|---|---|
| Latest release | **v1.3.9**, 2026-08-15 |
| Releases since the note | v1.3.1 through v1.3.9 — nine |
| Recent cadence | four releases between 2026-08-13 and 2026-08-15 |
| Last commit | 2026-08-17 |
| Recent contributors | several first-timers merged in the last week |

The reason I'm raising it: that note is the first thing a prospective user reads, and it is what search engines surface when people look up whether Shelfmark is still alive. It very nearly led us to migrate off Shelfmark and evaluate forks before we checked the commit history and found the project in better shape than the readme claims. I suspect we're not the only ones who have bounced off it.

Would you consider updating or removing it? Something that matches what you actually said in #1045 would be more accurate — for example that the project is maintained on a best-effort basis, with bug fixes and security updates handled and community PRs welcome, but no roadmap for new features right now.

Happy to open a PR with that wording if it would save you time — just say the word and I'll match whatever tone you prefer.

Either way, thanks for keeping this going. It's a genuinely useful piece of software.

calibrain:
Fair, I updated the message to be clearer
I dont have time to work on Features or big projects, but I still take some time every weekend to go through bug fixes and security issues

PR #1232, opened 2026-08-19
fix(bypass): scope browser cleanup to the calling session

### LINKED ISSUE #1231: [BUG] Internal bypasser kills concurrent workers' browsers (pkill -f chrome/Xvfb is container-wide)
### Describe the bug

When `MAX_CONCURRENT_DOWNLOADS > 1` and the **internal bypasser** is used, concurrent downloads destroy each other's browsers. Each worker that starts a bypass runs the orphan-process cleanup, which does a container-wide `pkill -9 -f chrome|chromium|Xvfb|ffmpeg`. That matches the browsers currently in use by the *other* workers, so their bypass attempts die with `Failed to connect to the browser`, and those downloads end as `All download sources failed`.

Since Anna's Archive now serves a DDoS-Guard challenge on essentially every request, almost every direct download needs the bypasser, so with the default `MAX_CONCURRENT_DOWNLOADS: 3` any batch of queued direct downloads reliably knocks itself over.

### Relevant code

`shelfmark/bypass/internal_bypasser.py`, orphan cleanup:

```python
processes_to_kill = ["chrome", "chromium", "Xvfb", "ffmpeg"]
...
kill_result = subprocess.run(
    [_PKILL_PATH, "-9", "-f", proc_name], ...
)
```

`pgrep`/`pkill -f` are unscoped, so they match every browser in the container, not just orphans belonging to the calling worker.

### To Reproduce

1. Internal bypasser enabled (`USE_CF_BYPASS: true`, `USING_EXTERNAL_BYPASSER: false`).
2. `MAX_CONCURRENT_DOWNLOADS: 3` (default).
3. Queue 3 direct downloads from Anna's Archive at once, so all three need a challenge solve.
4. Typically one succeeds; the others fail with `Failed to connect to the browser`.

### Logs

```
internal_bypasser.py:1114 - Pure CDP browser startup failed: Exception:
                --------------------------------
                Failed to connect to the browser
                --------------------------------
internal_bypasser.py:469 - Found 2 orphan chrome process(es), killing...
internal_bypasser.py:469 - Found 6 orphan chromium process(es), killing...
internal_bypasser.py:469 - Found 2 orphan Xvfb process(es), killing...
internal_bypasser.py:489 - Cleaned up 10 orphan process(es)
http.py:349 - Bypasser error: RuntimeError: RuntimeError: Pure CDP browser startup failed:
                Failed to connect to the browser
direct_download.py:1122 - Failed to fetch AA page: Unable to reach download source. Network restricted or mirrors are blocked.
```

Note the counts: those "orphans" are live browsers belonging to the other two workers.

### Workaround

Setting `MAX_CONCURRENT_DOWNLOADS: 1` fixes it completely. The same three downloads that failed as a concurrent batch all completed successfully, one after another, with no other change.

### Suggested fix

Scope the cleanup to the calling worker: track the PIDs/session the worker spawned and kill only those, or guard bypass sessions with a global lock so only one browser runs at a time. A cheaper mitigation would be skipping the sweep entirely while any other bypass session is active.

### Expected behavior

Concurrent downloads that each require a challenge solve should not terminate each other's browser processes.

### Version

- Shelfmark `v1.3.9`, build `2026-08-15-ebb833a82cb56ea50c28cfec41d1e220019ca075`
- `ghcr.io/calibrain/shelfmark:latest`, Docker on Linux
- Internal bypasser, custom DNS (manual: 1.1.1.1, 8.8.8.8, 9.9.9.9)

Unrelated to #1220 (the DDoS-Guard/cookie handling there is fixed for me on 1.3.9 — AA search returns results again). This one is purely the concurrency interaction.

calibrain:
The orphan sweep ran a container-wide 'pkill -9 -f chrome|chromium|Xvfb|ffmpeg', so it also matched browsers another bypass was still driving. Scope it by process group: kill only our own group and groups whose leader has died. Spawn the helper with start_new_session so its browser tree is identifiable, tear the whole group down after every run (a timed-out helper used to leak its Chrome and Xvfb), and have an orphaned helper take its browser down with it. Fixes #1231.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -3,7 +3,7 @@
 <img src="src/frontend/public/logo.png" alt="Shelfmark" width="200">
 
 > [!NOTE]
-> This project is in a stable state as of May 2026 but is not under active maintenance. 
+> Shelfmark is feature stable and maintained on a best-effort basis. Bug fixes, security updates, and small quality-of-life improvements are still shipped, and pull requests are reviewed — including new features. There is no roadmap for new features for now.
 
 Shelfmark is a self-hosted web interface for searching and requesting books and audiobooks across multiple sources. Bring your own sources, metadata providers, and download clients to build a single hub for your digital library. Supports multiple users with a built-in request system, so you can share your instance with others and let them browse and request books on their own.
 
@@ -238,9 +238,11 @@ These are non-goals, not missing features.
 
 ## Contributing
 
-Shelfmark's core feature set is complete. Development focuses on stability, bug fixes, quality-of-life improvements, and refining the search experience. Contributions in these areas are welcome, please file issues or submit pull requests on GitHub.
+Shelfmark's core feature set is complete.
 
-Feature requests that fall outside the project scope (library integration, automation, collection management) will be closed. If you're unsure whether something fits, open a discussion first.
+Pull requests are welcome and all of them get reviewed, new features included. If you want a feature, the fastest path is to send a PR for it rather than to file a request.
+
+Feature requests that fall outside the project scope (library integration, automation, collection management) will be closed, and PRs implementing them won't be merged. If you're unsure whether something fits, open a discussion first.
 
 ## Health Monitoring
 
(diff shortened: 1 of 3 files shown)


calibrain:
Also fixes #1225

PR #1237, opened 2026-08-20
fix: preserve multi-file audiobook folders

### LINKED ISSUE #1181: Multi-file audiobook downloads are not kept in their own folder when copied to the books destination
When I download a multi-file audiobook, such as an audiobook split into many mp3 chapter files, qBittorrent correctly saves it inside its own folder. For example, one download appears as a folder named "Lord Of The Rings (Complete Saga) - J. R. R. Tolkien" containing all of its chapter files together.

When Shelfmark copies this completed download into the books destination folder, **it does not keep these files together in that same folder.** Instead, it copies each individual file straight into the top level of the books folder. This means all the chapter files from every audiobook I have downloaded end up mixed together in one single flat folder.

This is a problem because many audiobook chapter files do not include the book title in the filename. Chapter files are often named things like "17 - The Clouds Burst.mp3" or "10 - A Warm Welcome.mp3" or simply a track number. Once these files are separated from their original folder and mixed in with files from other audiobooks, there is no way to tell which book a chapter belongs to just by looking at the file. It also causes Audiobookshelf to treat each chapter file as its own book.

mvanhorn:
Pass the effective source root from `process_folder_output` into `transfer_book_files`, and have the transfer layer select a sanitized child directory named after that source root when an audiobook has multiple files and its organization mode is `none` or `rename`. Create that grouping directory before applying the existing hardlink/copy/move logic so operation accounting, torrent seeding preservation, collision handling, cleanup, and custom-script final paths continue to use the established production path. Completed multi-file audiobook torrents arrive as a directory whose chapter filenames may not identify the book, but folder output currently sends every discovered chapter directly to the configured destination in `none` and `rename` modes. This flattens chapters from unrelated books together and causes directory-oriented consumers such as Audiobookshelf to interpret individual chapters as separate books.

A multi-file audiobook torrent in the default `rename` mode copies or hardlinks all supported chapter files beneath `<destination>/<original torrent directory>/` with their original chapter filenames, and places no chapters directly in the destination root; A multi-file audiobook in `none` mode receives the same source-folder grouping without renaming its chapter files.

Fixes #1181

### DIFF
--- a/shelfmark/config/settings.py
+++ b/shelfmark/config/settings.py
@@ -766,7 +766,10 @@ def _on_save_downloads(values: dict[str, Any]) -> dict[str, Any]:
             }
 
     # Audiobooks are always folder output.
-    if effective.get("FILE_ORGANIZATION_AUDIOBOOK", "rename") == "rename":
+    if effective.get("FILE_ORGANIZATION_AUDIOBOOK", "rename") in {
+        "rename",
+        "rename_and_group",
+    }:
         template = effective.get("TEMPLATE_AUDIOBOOK_RENAME", "")
         if _contains_path_separators(template):
             return {
@@ -1294,6 +1297,11 @@ def download_settings() -> list[SettingsField]:
                     "label": "Rename and Organize",
                     "description": "Create folders and rename files using a template. Recommended for Audiobookshelf. Do not use with ingest folders.",
                 },
+                {
+                    "value": "rename_and_group",
+                    "label": "Rename and Group",
+                    "description": "Rename single-file downloads; keep multi-file downloads grouped in their source folder.",
+                },
             ],
             default="rename",
             universal_only=True,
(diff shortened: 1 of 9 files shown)


calibrain (changes requested):
Thank you very much for the PR @mvanhorn!

If you don't mind, could you gate this behind a 4th select option? Something like `"Rename and Group"` in `FILE_ORGANIZATION_AUDIOBOOK` (`shelfmark/config/settings.py`), with the default staying on `Rename`.

Some people already have automation around the expected output, and I don't want to break their instance with an update.

Also, could you add support for usenet? `is_torrent_source` is `false` when it comes from usenet, but it can still be multi-file.

One last thing: `source_root` isn't always a directory — a single-file torrent holding an archive that extracts to multiple chapters would give you a `Book.zip/` folder. An `is_dir()` check should cover it.

Thanks again!

mvanhorn:
@calibrain thanks, all three covered in 6ca5a02.

**Rename and Group** is now a fourth `FILE_ORGANIZATION_AUDIOBOOK` option, and the default stays `Rename`. The grouping branch only runs when the mode is `rename_and_group`, so an instance that never changes the setting produces byte-identical output to today. The rename template shows for both `rename` and `rename_and_group`.

**Usenet**: dropped the `is_torrent` condition entirely, so the branch keys on `len(book_files) > 1` rather than the source. A multi-file usenet download groups the same way a torrent does.

**source_root**: added a `run_blocking_io(source_root.is_dir)` guard, so a single-file torrent whose archive extracts to multiple chapters no longer gets its `Book.zip` name used as the folder.

Covered by new cases in `tests/config/test_download_settings.py` and `tests/core/test_processing_integration.py`.

calibrain (approved):

calibrain:
Awesome, thank you !

PR #1242, opened 2026-08-20
fix(audiobookbay): search the ASCII punctuation ABB actually stores

calibrain:
WordPress texturizes punctuation on output only, so a post stored as "The
Stranger's Wife" renders as "The Stranger’s Wife". ABB's search matches the
stored value and ANDs its terms, so one typographic character in the query
empties the entire result set rather than merely ranking worse. Book metadata
and mobile keyboards both hand us those characters.

Map curly quotes, dashes and ellipses to ASCII before a query goes out, and on
both sides of the relevance comparison, since scraped titles carry the rendered
forms. Release titles are still stored and displayed exactly as ABB renders
them; only matching normalizes.

Also percent-encode the search query properly. The hand-rolled encoder only
escaped double quotes and spaces, so a bare "&" started a new query parameter
and silently truncated the search: "detective dan riley books 1 & 2 weatherley"
reached ABB as "detective dan riley books 1" and returned six confident-looking
results without the requested book among them. "%" and "+" were mangled too.

### DIFF
--- a/shelfmark/release_sources/audiobookbay/scraper.py
+++ b/shelfmark/release_sources/audiobookbay/scraper.py
@@ -2,7 +2,7 @@
 
 import re
 import time
-from urllib.parse import quote
+from urllib.parse import quote, quote_plus
 
 import requests
 from bs4 import BeautifulSoup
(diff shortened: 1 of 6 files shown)


calibrain:
Fix #1194

PR #1255, opened 2026-08-22
feat(search): let each user pick their own default book languages

helgehelge123:
## Why

`BOOK_LANGUAGE` is a per-reader property, not a per-instance one. On a shared install one household member searches in German while another wants English and German — today whoever changes the setting changes it for everyone, and the only escape is re-picking languages in the filter on every single search.

The per-user override machinery already carries `SEARCH_MODE`, the metadata providers and the default release sources, so the language default mostly had to opt into it.

## What changed

**The field.** `BOOK_LANGUAGE` becomes `user_overridable` and moves from the **General** tab to **Search Mode**, next to the other user-overridable search defaults (per [review](https://github.com/calibrain/shelfmark/pull/1255#issuecomment-5391189094) — the first version had the Search section span two tabs, this one doesn't). Admins set it per user in the user editor, users set it in **My Account → Search Preferences**, and the Search Mode tab carries the usual "N users override this" summary.

**No migration for the move.** `general` and `search_mode` both persist into `settings.json`, and a field's value is resolved through `load_config_file(tab)` for the tab it's declared on — so an install that already stores `BOOK_LANGUAGE` keeps its value. Checked against a `settings.json` written while the field still lived on General: the stored value resolves unchanged, a fresh install still gets `["en"]`, and `BOOK_LANGUAGE` in the environment still overrides both.

**The two places the default is read.**

- `/api/config` seeds the frontend's language filter, so it now resolves `BOOK_LANGUAGE` for the session user.
- `build_release_search_plan` falls back to the default whenever a request carries no language filter — which is exactly what the filter's "Default" option sends. It takes an optional `user_id`, passed by `/api/releases` from the session and by the Prowlarr retry path from `task.user_id`, so a retry re-searches in the languages of whoever queued the download.

**Validation.** Overrides go through `normalize_language()`, so `"German"`, `"ger"` and `"de"` all store as `de`, and an unknown language is rejected with a message naming it instead of being silently searched for. An empty list stays an empty list (a deliberate "no default filter"), `null` clears the override as everywhere else, and ENV still wins: with `BOOK_LANGUAGE` set in the environment the field reports `fromEnv` and overrides are ignored.

**Scope.** Only the language default becomes overridable. The two format lists left behind under "Default Search Filters" stay admin-only — they describe what the library and its post-processing accept, not what a reader wants to read. There's a test pinning that.

## Verification

- 2681 unit tests pass (2670 before, 11 added)
- `ruff check`, `ruff format`, `basedpyright` over backend and tests, and `vulture` all clean; frontend lint, format, typecheck and 126 unit tests clean
- `docs/environment-variables.md` regenerated via `scripts/generate_env_docs.py` (the `BOOK_LANGUAGE` row follows the field into the Search Mode section)
- Manually against a two-user instance with builtin auth (first round, before the tab move): with user A on German and user B on English+German, `/api/config` returns each reader their own `default_language` and an unfiltered `/api/releases` plans the matching languages; an admin can set and read the same override for another user; clearing it falls back to the global value; a stray `"klingon"` is rejected; and `BOOK_LANGUAGE` in the environment overrides both users with the field marked `fromEnv`
- After the tab move I re-ran the suites above plus the stored-value/fresh-install/ENV check described under "No migration for the move"; the behaviour it exercises is what the move could have broken

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -248,7 +248,6 @@ Seconds since the last WireGuard handshake before the healthcheck bounces the tu
 | `AUDIOBOOK_LIBRARY_URL` | Adds a separate navigation button for your audiobook library (Audiobookshelf, Plex, etc). When both URLs are set, icons are shown instead of text. | string | _none_ |
 | `SUPPORTED_FORMATS` | Book formats to include in search results. ZIP/RAR archives are extracted automatically and book files are used if found. | string (comma-separated) | `epub,mobi,azw3,fb2,djvu,cbz,cbr` |
 | `SUPPORTED_AUDIOBOOK_FORMATS` | Audiobook formats to include in search results. ZIP/RAR archives are extracted automatically and audiobook files are used if found. | string (comma-separated) | `m4b,mp3,m4a,flac,ogg,wma,aac,wav,opus,zip,rar` |
-| `BOOK_LANGUAGE` | Default language filter for searches. | string (comma-separated) | `en` |
 
 <details>
 <summary>Detailed descriptions</summary>
@@ -298,15 +297,6 @@ Audiobook formats to include in search results. ZIP/RAR archives are extracted a
 - **Type:** string (comma-separated)
 - **Default:** `m4b,mp3,m4a,flac,ogg,wma,aac,wav,opus,zip,rar`
 
-#### `BOOK_LANGUAGE`
-
-**Default Book Languages**
-
-Default language filter for searches.
-
-- **Type:** string (comma-separated)
-- **Default:** `en`
-
 </details>
 
 ## Search Mode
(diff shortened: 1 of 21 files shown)


calibrain:
Hi ! Yeah, moving it to the Search Mode tab instead would be ideal if you dont mind !

Otherwise it looks great, thank you !

helgehelge123:
Done — `BOOK_LANGUAGE` now lives on the **Search Mode** tab, right under `SEARCH_MODE`, and the cross-tab payload machinery is gone: `build_user_preferences_payload` and `self_user_routes.py` are back to exactly what they were on `main`, so the Search section maps onto one tab again. The whole diff in `shelfmark/core/` is a single added line now (`BOOK_LANGUAGE` in the search-preference key set in `admin_settings_routes.py`).

No migration needed, and I checked that rather than assuming it: `general` and `search_mode` both persist into `settings.json`, and a field's value is resolved via `load_config_file()` for the tab it's declared on. Against a `settings.json` written while the field still lived on General, the stored value resolves unchanged (`['de', 'hu']`), a fresh install still gets `['en']`, and `BOOK_LANGUAGE` in the environment still overrides both.

Also rebased onto current `main` (the branch was 6 commits behind) and regenerated `docs/environment-variables.md`, which moves the `BOOK_LANGUAGE` row into the Search Mode section.

Tests: the settings-definition test for the field moved from `tests/config/test_general_settings.py` to `test_search_mode_settings.py`; `test_general_settings.py` keeps only the pin that the two format lists stay admin-only; the expected `keys` order in `TestAdminSearchPreferences` now has `BOOK_LANGUAGE` second, following the tab order; and the test that pinned the two-tab payload is now a single-tab one. 2681 unit tests pass (2670 before, 11 added), with `ruff`, `ruff format`, `basedpyright`, `vulture` and the frontend lint/format/typecheck/unit tests clean.

Thanks for the review!

PR #1259, opened 2026-08-23
Integrations/cwa sidecar auto release

EdLaughton:


calibrain:
Hi ! Can you add a PR description on what this does and whats the use case ?
(also, did cwa start supporting metadata files ? That sounds awesome !)

downey-lv:
I might be missing something, but from the ingest code on CWA it looks like that only add_format action is supported from the ingest folder and this would have no effect, the json would just stay in the ingest folder

https://github.com/crocodilestick/Calibre-Web-Automated/blob/43718d85f6b0adbc2faff36c393fc7a823c60fd7/scripts/ingest_processor.py#L1424

calibrain:
@EdLaughton will convert to draft for now until further details (also this PR drifted significantly and needs some merge conflict and rebasing)

PR #1269, opened 2026-08-26
feat(homepage): Always show controls

ndawg:
I found using the main search menu frustrating. Often times, I would type in what I want, then select the category, only for it to get erased. And the menu closing over and over was distracting. So this PR makes the buttons stick around permanently and it removes the search field text changing with each button press.

Obviously, this is just what I want, but I figured I'd at least put a PR up for it.

image
image


calibrain (changes requested):
Hey thank you, I love the idea !
But to not completely change the current behaviour, can I ask for a change ? 
instead, can we, when we click on the search button 
image

Have it "lock" the menu (as in keep it always visible, even make it flat like your PR does, I love it)
And if we click on it again, it goes back to hiding ?

joshft91:
I like the direction this PR is headed. I'm wondering if behavior between the two above would more intuitively align with users:

Instead of having to click a link to keep it around like @calibrain mentioned, what about allowing the user to keep the dialogue box open/select-able while the mouse is over the area, then closing when the mouse leaves the area. Here's an example from a now-abandoned fork of Shelfmark, and I'm looking to come back to this repo as it's more actively maintained.

https://github.com/user-attachments/assets/36442658-d79a-408b-8b82-b40eb0ab8f9f

calibrain:
@joshft91  yes, that works great, I love the idea !

ndawg:
@calibrain I kept it so that you have to click the button first, as I didn't see you reply to josh before I did it. Let me know if you want we he did instead!

calibrain (approved):
Looks great !

hobesman:
@calibrain One issue is that if you click to enable "Options", the options settings jump out from under your cursor as soon as you try to click on them, because clicking away from the search menu makes it disappear and moves the options up on the page. Can we either include the options as part of the search menu for purposes of keeping the search menu visible, or make the options stay put when the search menu disappears?

Should I make a separate issue for the continued discussion?

calibrain:
@hobesman can you please create a new issue with a screenshot ?
@ndawg if you have itime, can you take a look at this ?

If not, I should have some time next week 
Thank you !

PR #1282, opened 2026-08-28
fix(qbittorrent): keep magnets whose metadata is still pending

jfmlima:
## Problem

`QBittorrentClient.add_download()` waits 20 × 0.5 s for qBittorrent to leave `metaDL`, then raises:

```
Failed to add to qbittorrent: Torrent metadata resolution was not confirmed within the visibility grace period
(response=TorrentsAddedMetadata({'added_torrent_ids': [], 'failure_count': 0, 'pending_count': 1, 'success_count': 0}))
```

The wait exists to learn qBittorrent's primary torrent ID, which for hybrid torrents switches from the v1 hash to the truncated v2 hash once metadata resolves. A magnet on a thin public swarm routinely needs longer than 10 s to find a peer that will serve metadata, and the download is then abandoned even though the add itself succeeded. The torrent stays in qBittorrent (`base_handler` logs "leaving in qbittorrent") and often completes minutes later with nobody watching it.

Seen on v1.3.12 with public indexers through Prowlarr: every magnet-only release failed this way, while `.torrent` releases from a private indexer were fine. qBittorrent showed the same torrents at `metaDL 0% seeds=0/0`, and they resolved on their own well after shelfmark had given up.

## Change

Return the info hash we already have instead of raising when the grace period expires. Reads then resolve either identity:

- `get_status()` and `get_download_path()` use `_resolve_torrent()` instead of `_get_torrent_info()`, so a v1 hash still matches after qBittorrent re-keys the torrent to v2. `_torrent_matches_download_id` already compares `hash`, `infohash_v1` and `infohash_v2`.
- `remove()` and `set_category()` address the torrent by its current primary hash through a new `_current_hash()` helper, which falls back to the ID it was given when the torrent cannot be resolved.
- The two magic numbers become `_METADATA_WAIT_POLLS` and `_METADATA_WAIT_INTERVAL_SECONDS`.

The happy path does not change. When metadata resolves inside the grace period the resolved primary hash comes back as before, and `_resolve_torrent()` tries the exact-hash lookup first, so it costs no extra request.

## Tests

`test_add_fails_when_metadata_never_resolves` asserted the old behaviour, so it becomes `test_add_keeps_torrent_when_metadata_never_resolves` and asserts the info hash is returned. `test_get_status_resolves_hash_after_metadata_switch` is new: it reads status by the v1 hash after qBittorrent reports the torrent under its v2 hash.

`uv run pytest tests/ --ignore=tests/e2e` gives the same 55 failures with and without this change (they are all in `tests/bypass/` and need Chrome, which my machine has no headless setup for), and `tests/prowlarr/` is green at 524 passed. Ruff check and format are clean. I have not run this branch against a live qBittorrent, so a second pair of eyes on the `remove()` path would help.

### DIFF
--- a/shelfmark/download/clients/qbittorrent.py
+++ b/shelfmark/download/clients/qbittorrent.py
@@ -45,6 +45,10 @@ _HASH_LENGTH_ED2K = 32
 _HTTP_STATUS_FORBIDDEN = HTTPStatus.FORBIDDEN
 _HTTP_STATUS_NOT_FOUND = HTTPStatus.NOT_FOUND
 _METADATA_DOWNLOAD_STATES = {"forcedMetaDL", "metaDL"}
+# How long add_download waits for magnet metadata before falling back to the info
+# hash it already knows, rather than holding the download queue on a thin swarm.
+_METADATA_WAIT_POLLS = 20
+_METADATA_WAIT_INTERVAL_SECONDS = 0.5
 _ONE_WEEK_IN_SECONDS = 604800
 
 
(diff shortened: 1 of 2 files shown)


calibrain:
Good catch, thank you !

Issue #1276, opened 2026-08-27
[BUG] Releases fail to display / Load


Ditaveloci:
## Bug Description
This may be linked to other reported bugs so i apologize if so. 
On selection of item, releases page attempt to load and ulitmately failes with the 

## Steps To Reproduce

1. Search book
2. Click book 
3. wait for fail to load releases " Error Loading Releases, Server unavailable (504). If using a reverse proxy, check its configuration."

## Expected Behavior

Releases load, and can be downloaded.

## Debug Information
**Important**: Please enable debug mode before submitting a bug report by setting the environment variable:
```
DEBUG=true
```

When running in debug mode, a DEBUG button will appear in the interface. Please click this button to generate a debug zip file and attach it to this issue.

## Environment Information

- BASE OS: [e.g. Windows 10, Ubuntu 22.04]
Ubuntu 24.04 

## docker-compose

  shelfmark:
    image: ghcr.io/calibrain/shelfmark:latest
    container_name: shelfmark
    healthcheck:
      disable: true
    environment:
      CUSTOM_DNS: cloudflare
      USE_DOH: true
      FLASK_PORT: 8084
      LOG_LEVEL: info
      BOOK_LANGUAGE: english
      USE_BOOK_TITLE: true
      TZ: Europe/London
      APP_ENV: prod
      UID: 1000
      GID: 1000
      CWA_DB_PATH: /auth/app.db
      INGEST_DIR: /cwa-book-ingest
      DEBUG: true
    volumes:
      - Redacted 
    network_mode: "service:protonvpn"
    restart: unless-stopped
    depends_on:
      - protonvpn

[shelfmark-debug_BUILD-2026-08-24-89104ae80f1019c778aee648abe7b5ee672e94f8_RELEASE-v1.3.12_20260827-134324.zip](https://github.com/user-attachments/files/31515124/shelfmark-debug_BUILD-2026-08-24-89104ae80f1019c778aee648abe7b5ee672e94f8_RELEASE-v1.3.12_20260827-134324.zip)

jfmlima:
Same symptom here on v1.3.12 (BUILD 2026-08-24-89104ae). The internal bypasser stopped solving DDoS-Guard on AA `/search` between the evening of 2026-08-27 and the morning of 2026-08-28, on the same egress IP.

[...]

PR #1286, opened 2026-08-30
fix(bypass): wait for the solved page before reading its source

jfmlima:
Follow-up to #1276 with a measurement from the instance I reported there. v1.3.13 solves the challenge again, but on my setup the solve was being thrown away immediately afterwards:

```
19:26:08 Bypass successful using _bypass_method_cdp_gui_click
19:26:16 Bypass failed (attempt 1/10): TimeoutError: Time ran out while waiting for: {html}
```

`_get()` ends with `return await page.get_page_source()`, which is `find("html", timeout=1)` in SeleniumBase. One second is enough for a page that is already sitting on its content, but Anna's Archive answers a cleared check with a redirect to the real page, so the document is not there yet. The solve is discarded, the whole attempt restarts, and the extra requests are what earn the 429 that `note_rate_limited()` then parks the host for — 120 s, then 300 s.

## Change

`_read_page_source()` waits for the document itself, with a `BYPASS_PAGE_SOURCE_TIMEOUT` setting (default 20 s, min 1, max 120) in Direct Download → Cloudflare Bypass, next to the existing bypasser timeouts.

## Measured on a live instance

I patched the wait in the running container (`find("html", timeout=1)` → `timeout=20` in the installed seleniumbase, which is the same effect as this PR) and re-ran the same searches on the same host, k3s behind a Surfshark WireGuard exit, internal bypasser, v1.3.13:

| | 1 s wait | 20 s wait |
|---|---|---|
| `Time ran out while waiting for: {html}` | one per solve | none |
| 429 backoffs | 2 (120 s, then 300 s) | none |
| Search for a book AA has | 199 s and 200 s, both errored | 61 s, 2 epub releases |

A download after that took 5 s from LibGen, so the search was the whole cost.

## Tests

Two tests in `tests/bypass/test_bypass_budgets.py`, the file already covering #1276: a page that needs longer than a second still yields its HTML, and `BYPASS_PAGE_SOURCE_TIMEOUT` overrides the default.

`uv run pytest tests/ --ignore=tests/e2e`: 2848 passed, 47 skipped. Ruff check and format clean. The docs table is auto-generated, but running `scripts/generate_env_docs.py` here rewrote unrelated entries (Newznab, BOOK_LANGUAGE), so I added only the new entry by hand in the generator's format rather than commit that churn.

One thing I could not judge from outside: whether 20 s is the right default for hosts other than AA. It only costs anything when a solve would otherwise be discarded, but I have measured it on one site.

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -2344,6 +2344,7 @@ Override destination based on content type metadata.
 | `EXT_BYPASSER_URL` | URL of the external bypasser service (e.g., FlareSolverr). | string | `http://flaresolverr:8191` |
 | `EXT_BYPASSER_PATH` | API path for the external bypasser. | string | `/v1` |
 | `EXT_BYPASSER_TIMEOUT` | Timeout for external bypasser requests in milliseconds. | number | `60000` |
+| `BYPASS_PAGE_SOURCE_TIMEOUT` | How long to wait for a solved page to produce its content before the bypass is retried. Raise it if solves succeed but searches still fail. | number | `20` |
 | `BYPASS_BROWSER_IDLE_TIMEOUT` | How long the bypass helper process may sit unused before it is shut down. Higher keeps more searches fast, lower frees memory sooner. | number | `180` |
 
 <details>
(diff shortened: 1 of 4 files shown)


calibrain:
Thank you so much

jfmlima:
@calibrain you're missing the release 😄

calibrain:
@jfmlima working on some other bug fixes and will bundle them together today or tomorrow !

jfmlima:
Ah, sounds good, thanks!

PR #1309, opened 2026-09-04
fix(postprocess): attach unmatched chaptered audio files to existing book group (#1176)

### LINKED ISSUE #1176: [BUG] Auto-Organize fails with big number of .flac files creating multiple folders
## Bug Description
I found an audiobook through an Indexer (Prowlarr). It was about 3,2GB consisting of about 254 .flac files after unpack. 
The download with SABnzbd went without problems after I turned on "+Delete" as donwload option so there were no .rar files causing issues in the downloads folder. I also allowed .flac files in the compose.yaml without issues so far. 
But while extracting and importing I checked the folders it generated: 
1. It started with a folder named "Westwell - Hot & Cold (2023) " putting .flac files there (as expected).
2. After about 110 .flac files it created a new folder called "Hot & Cold" and started putting the remaining .flac files in this folder. 
This resulted in two seperate folders, which then were considered to be two seperate audiobooks in Audiobookshelf.
I tried multiple time also with different but similar files (also big amounts of .flac) encountering the same issue most of the time, not always.
In the UI it showed an error with the download. Sadly I don't have an exact screenshot, but it was something like "file not found". Renaming of files went perfectly all the way. 

## Steps To Reproduce
0. Indexer in Prowlarr, SABnzbd with +Delete as default as downloader, .flac files allowed for audiobooks
1. Start +Get with big collection of .flac files (" *-16BIT-44.1KHZ-WEB-FLAC-2023- ")
2. Check files folder while import is running

## Expected Behavior
I expected all .flac files to be renamed and be moved in the same folder. 

## Environment Information
- BASE OS: Multiple Debian VMs with Docker connected through NFS on the same HDD. (Linux 7.x - 2.6 Kernel)

## docker-compose
image: ghcr.io/calibrain/shelfmark-lite:latest
additionally: "SUPPORTED_AUDIOBOOK_FORMATS: m4b,mp3,m4a,flac"

amasen02:
### Summary
Fixes #1176

When downloading an audiobook with many chaptered tracks (e.g. 250+ `.flac` or `.mp3` files), indexer XML or release metadata often caps the file list at ~100-110 entries. When the release extracts on disk, `match_plan_to_files()` matched those first ~110 files to the planned book group, while the remaining 140+ files fell into `unmatched` and triggered fallback heuristic grouping.

Because heuristic grouping parsed the folder name (`Westwell - Hot & Cold (2023)`) and stripped the series/author prefix, it generated a second book titled `Hot & Cold` containing the remaining tracks, resulting in two split book folders.

### Changes
- In `match_plan_to_files()` (`shelfmark/download/postprocess/packs.py`), check `unmatched` files before falling back to heuristic multi-book splitting.
- If an unmatched file is chaptered audio (`.flac`, `.mp3`, `.aac`, etc.) and shares the directory with an existing book group, or if the plan was a single-book plan, append it to that group instead of creating a secondary book.
- Non-chaptered standalone books (e.g. `.m4b`, `.epub`) or files in separate subfolders continue to fall back to heuristic grouping as before.
- Added unit tests in `tests/download/test_packs.py` verifying:
  1. Truncated track list in single folder properly appends remaining chaptered tracks without splitting.
  2. Single-book plan with multi-disc audio files (`CD1`/`CD2`) groups together cleanly.
  3. Multi-book packs with unmatched chaptered tracks route each track to its respective book folder.

### Testing
Ran `uv run pytest tests/download/test_packs.py` and `uv run pytest tests/core/test_processing_packs.py` (all passed cleanly). Checked type annotations with `basedpyright` (0 errors) and formatting with `ruff`.

### DIFF
--- a/shelfmark/download/postprocess/packs.py
+++ b/shelfmark/download/postprocess/packs.py
@@ -414,10 +414,29 @@ def match_plan_to_files(
             )
 
     unmatched = [p for p in book_files if p not in claimed]
-    if unmatched:
+    still_unmatched: list[Path] = []
+    for p in unmatched:
+        p_ext = p.suffix.lower().lstrip(".")
+        is_chaptered = p_ext in _CHAPTERED_AUDIO_EXTENSIONS
+
+        matching_groups = [g for g in groups if any(f.parent == p.parent for f in g.files)]
+
+        target_group: BookGroup | None = None
+        if len(matching_groups) == 1 and is_chaptered:
+            target_group = matching_groups[0]
+        elif len(plan) == 1 and len(groups) == 1 and is_chaptered:
+            target_group = groups[0]
+
+        if target_group is not None:
+            target_group.files.append(p)
+            claimed.add(p)
+        else:
+            still_unmatched.append(p)
+
+    if still_unmatched:
         groups.extend(
             group_files_into_books(
-                unmatched, series_name=series_name, author_name=author_name, root=root
+                still_unmatched, series_name=series_name, author_name=author_name, root=root
             )
         )
     return groups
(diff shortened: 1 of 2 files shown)


calibrain:
Thank you !

PR #1314, opened 2026-09-05
Keep default filters out of the URL hash

nfvelten:
Follow up to #1311, per your "no use keeping empty / default values in the url".

A plain author search was coming out as

```
#q=herbert&search_by=author&sort=relevance&lang=default&format=epub&format=mobi&format=azw3&format=fb2&format=djvu&format=cbz&format=cbr
```

The builder serialized every advanced filter regardless of whether the user had touched it. Now `sort` drops when it equals the sort the app would apply anyway (the provider default in Universal, the configured one in Direct), `lang` drops while it is still the `default` sentinel, and `format` drops when the selection matches `supported_formats` from the config. Formats are compared as a set, so reselecting everything in a different order still counts as default. Anything the user actually changed stays.

That also makes the comment I left in the builder true: a default-state URL now carries no hash at all.

The second commit is an ordering bug that omitting defaults made visible. `loadConfig`'s `initial` branch seeds `formats` from `supported_formats` and `sort` from the mode default. The URL bootstrap is gated on config being loaded, so it normally runs after that and wins on its own, but nothing guarantees only one `initial` load happens, and a second one landing after the bootstrap resets `formats` to the full supported list and drops the sort the link asked for. StrictMode double-invokes the mount effect that triggers it, so it reproduces in development: `#q=dune&format=epub&lang=en` intermittently loses its `format=epub`. The seeding is now skipped once the bootstrap has applied, so a link's filters win over the defaults they were meant to override.

Five new unit tests on the builder. I also drove both the dev server and a production build in Chromium with `/api` mocked: `#search_by=author&q=herbert` settles at `#q=herbert&search_by=author`, typing a plain query gives `#q=dune`, direct mode at rest carries no hash, and `#q=dune&format=epub&lang=en` keeps both filters in the hash and in the request. That last one was 3 for 5 on the dev server before the second commit and 5 for 5 after.

### DIFF
--- a/src/frontend/src/App.tsx
+++ b/src/frontend/src/App.tsx
@@ -2012,6 +2012,13 @@ function App() {
     return searchFieldValues[activeQueryOption.field.key] ?? '';
   }, [activeQueryOption, searchInput, searchFieldValues]);
 
+  // The sort the app applies with no user choice, mirroring what loadConfig seeds
+  // advancedFilters.sort with - a sort equal to it is a default, not a shared intent.
+  const urlHashDefaultSort =
+    effectiveSearchMode === 'universal'
+      ? resolvedMetadataDefaultSort
+      : config?.default_sort || 'relevance';
+
   // Keep the URL hash fragment live as search state changes. Gated until any URL-driven
   // bootstrap has applied (or there was nothing to apply), so we don't clobber a shared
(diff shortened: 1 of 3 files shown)


calibrain:
Thank you very much !

PR #1326, opened 2026-09-10
feat(sources): add Libgen as a direct catalogue search source

klaidliadon:
## What

Adds **Libgen as a search source**. Today Libgen is only a download mirror (reached by an Anna's Archive md5), so anything in Libgen but not in AA's search index is invisible — and that's where most of the CBZ/CBR comics and manga live. A Libgen search for *One Piece*, for instance, turns up ~99 volumes that AA search never shows.

It's a self-contained `release_sources/libgen/` package (source + handler + settings) plus one line to register it. **No changes to `direct_download.py`** — it reuses the existing `ads.php → get.php` resolution and the mirrors already configured in `LIBGEN_MIRROR_URLS`. Plain HTTP, no bypasser needed (libgen.li isn't behind DDoS-Guard). Opt-in via a settings toggle.

## Worth a look in review

- **`source_id` is `libgen:<md5>`, not the bare md5.** The download queue keys on `task_id` (= `source_id`), and `direct_download` already uses the bare md5. Since AA indexes a lot of Libgen, the same md5 shows up from both sources — a bare id would collide in the queue. The handler strips the prefix before downloading.
- **Reachable like the other non-default sources** (Prowlarr, AudiobookBay, …): it appears in the per-book release search, not the free-text box (that stays wired to `direct_download`).

Tests in `tests/libgen/` cover parsing (both row layouts), the source, the handler, and `get_record`. Lint/format/typecheck clean.

### DIFF
--- a/shelfmark/release_sources/__init__.py
+++ b/shelfmark/release_sources/__init__.py
@@ -422,6 +422,7 @@ _BUILTIN_SOURCE_MODULES = (
     "shelfmark.release_sources.audiobookbay",
     "shelfmark.release_sources.direct_download",
     "shelfmark.release_sources.irc",
+    "shelfmark.release_sources.libgen",
     "shelfmark.release_sources.newznab",
     "shelfmark.release_sources.prowlarr",
 )
--- a/shelfmark/release_sources/libgen/__init__.py
+++ b/shelfmark/release_sources/libgen/__init__.py
@@ -1,0 +1,6 @@
+"""Libgen release source - direct catalogue search over the libgen.li family."""
+
+# Import to trigger registration
+from shelfmark.release_sources.libgen import handler as handler
+from shelfmark.release_sources.libgen import settings as settings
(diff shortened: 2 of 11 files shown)


klaidliadon:
Gentle nudge on this one. Quick recap to keep review cheap:

* Self contained `release_sources/libgen/` package (source, handler, settings), plus one line to register it
* No changes to `direct_download.py`
* Reuses the existing `ads.php` → `get.php` resolution and the `LIBGEN_MIRROR_URLS` mirrors
* Off by default behind a settings toggle
* Tests cover both row layouts, the source, the handler, and `get_record`

Since this is my first PR here the CI hasn't run. Could you approve the workflow so the suite can validate it? Happy to fix anything it flags.

calibrain:
Hey sorry, I am quite bus, so I dont have lots of time to review and merge code, but this looks great, thank you very much for the contribution !

PR #1333, opened 2026-09-11
Add configurable word separator for naming templates

### LINKED ISSUE #1230: [Feature Request] Provide option to use dots as word separator when renaming files
Hello,

The current file management options use space as a word separator.
For various reasons, a user might want to avoid spaces and use dots instead.

The *arr stack already offers this feature in the templating by using dotted keywords in the templates (for instance {original title} vs {orginal.title}

It would be nice if Shelfmark could offer something similar:
1. either by using the same approach with dotted keywords in the templates (but this might be complicated as currently placeholders only have one word), or by adding a trailing dot to the variable, ie `{Author.}` would give `Arthur.Conan.Doyle`
2. or have another option in the renaming template for the user to pick word separator from a a list ("space", dot, underscore, hyphen, eventually even allow the user to use a custom separator if they want to use something different.

I tried using this template for instance, but the outcome isn't great:
`{Author}/{Author}.-.{Title}.({Year})`
gave:
`Arthur Conan Doyle/Arthur Conan Doyle.-.The Hound of the Baskervilles_ Another Adventure of Sherlock Holmes.(1902).epub`
 (BTW I don't know where this underscore comes from in the preview, might be a bug/typo? )

Hope this makes sense, thanks !

viniciuspx:
Closes #1230

## What
Adds a "Word Separator" setting (Space / Dot / Underscore / Hyphen / Custom) that replaces internal whitespace in each naming-template placeholder's rendered value — e.g. `{Author}` renders "Arthur.Conan.Doyle" instead of "Arthur Conan Doyle" when Dot is selected.

This follows option 2 from the issue rather than inventing new dotted-keyword template syntax (`{Author.}`), since it's a smaller surface: one setting applies uniformly across all four templates (books/audiobooks × rename/organize) instead of needing a parallel token for every existing one.

## How it works
- Literal characters typed into the template itself (e.g. the `.` in `{Author}.-.{Title}`) are never touched — only whitespace *inside* a placeholder's resolved value is affected.
- Default is "Space", which is a no-op: existing templates produce byte-identical output after this change (verified via the existing test suite, unmodified, still passing).

## Where
- `shelfmark/core/naming.py` — `word_separator` param on `parse_naming_template` / `build_library_path`.
- `shelfmark/download/postprocess/policy.py` — `get_word_separator()`, mirroring the existing `get_file_organization()` accessor.
- `shelfmark/download/postprocess/transfer.py` — wires the resolved separator through the four existing template-rendering call sites.
- `shelfmark/config/settings.py` — new `Word Separator` / `Custom Word Separator` fields next to the existing naming-template fields.
- `src/frontend/.../namingTemplatePreview.ts` + `NamingTemplateField.tsx` — the settings UI has its own TS mirror of the Python renderer for the live preview; updated it in lockstep so the preview doesn't lie about what the separator will actually do.
- Tests added on both sides (pytest + vitest).

## Testing
- `uv run pytest tests/core/test_naming.py tests/core/test_destination_file_organization.py` — all pass, including new cases.
- `uv run pytest` (full suite) — same pre-existing failures as on `main` before this change (browser/network-dependent bypass & e2e tests unrelated to this diff), everything else green.
- `uv run ruff check` / `ruff format --check` / `basedpyright` — clean.
- `npm run lint` / `format:check` / `typecheck` / `test:unit` (196 tests) — clean.


calibrain:
Looks good, just one thing

Can we merge 
 `NAMING_WORD_SEPARATOR`
 `NAMING_WORD_SEPARATOR_CUSTOM`

Instead of `dot`, `underscore` etc. just set the character directly `.`, `_`, etc.
(and treat empty as space ` `) ?

viniciuspx:
Done — merged into a single `NAMING_WORD_SEPARATOR` setting that takes the character directly (e.g. `.`, `_`), with empty treated as space. Dropped `NAMING_WORD_SEPARATOR_CUSTOM` and the enum mapping accordingly, updated the frontend preview mirror, docs, and tests.

calibrain:
Thank you very much !

PR #1336, opened 2026-09-13
Feature: Add Download counts to search result displays and Download sidebar

RoninTech:
Having the download counts from AA as an input on choosing which of the many search results to pick for downloading is useful.  This PR makes the downloads numbers available on all the search result pages and also on the Download sidebar after the user presses a download button for a Direct Download.

I have also included a SKILL.md and associated download_books.py that can be used just as a reference or with an LLM harness for automation.  The Downloads info is used by the script to pick which search result to download out of the many available.

Since a picture is worth a thousand words:

search-results-with-downloads

download-sidebar-with-downloads

Coded with llama.cpp and 🤖


calibrain:
I love this idea ! Its awesome !
I merged https://github.com/calibrain/shelfmark/pull/1337 which did a few backend changes and it created some conflicts, could you fix them and I will approve this after ! 
Thank you !

RoninTech:
Should be good to go now.

RoninTech:
Let me know if you need any other info to allow this in. I really don't want to re-merge this again. 🙂

calibrain:
Awesome work, thank you so much !

PR #1337, opened 2026-09-13
refactor: make direct download provider-driven

TomJansen:
This is the refactor for the download handler


calibrain:
Thank you for splitting the PR, this looks mostly good !

PR #1343, opened 2026-09-17
Extract archives when zip/rar are enabled as supported formats

funkypenguin:
The default audiobook formats include `zip` and `rar`. `scan_directory_tree` checks the supported-format list before checking for archives, so a downloaded archive lands in `book_files` and is imported as-is. The extraction branch in `collect_directory_files` is never reached.

This keeps archives out of `book_files`, so they always take the archive path: extracted when extraction is allowed, imported as-is when it isn't (unchanged).

Tests added in `tests/download/test_postprocess_scan_archives.py`; three of the four fail without the change.


calibrain:
Good catch, thank you !

PR #1351, opened 2026-09-18
Added the ability to sort direct search results by Most downloads

RoninTech:
Now that we know about the AA Downloads stats for search results (#1336), this PR allows sorting them locally by "Most downloads" without doing another AA search.  Not needing to do an AA search for this is faster than the other sorts which do new AA searches asking for their results to be sorted a certain way.  Also added the "Settings->Search Mode->Default Sort Order->Most downloads" option.  Universal downloads do not allow for results sorting so I did not include this feature there.

NOTE: While adding this feature I noticed that shelfmark only looks at the first page of AA search results.  So switching between different sort orders may result in different sets of results with the same search term. 

Here's a couple of pics showing the new sort option:

search-mode-sort-order-most-downloads

direct-search-results-most-downloads


calibrain:
Looks good, thank you !

PR #1318, opened 2026-09-06
feat: add extensible direct-download providers and OceanofPDF support

TomJansen:
This pull request refactors direct downloads into an extensible, provider-based system, which allows Shelfmark to support sources beyond Anna’s Archive.

It also adds OceanofPDF as a direct-download provider (#1149), and includes mirror selection and onboarding support.

It also fixes the CI workflow by converting GitHub Container Registry image names to lowercase. This prevents publishing failures when a GitHub username contains uppercase letters.

Please note that I made this mostly with AI-assistance, and I understand that is a rather large pull request. I tested it and it works, though.

PR #1047, opened 2026-06-06
feat: optional Hardcover wishlist sync with automated, library-aware downloading

InfiniteAvenger:
## Summary
Adds an **opt-in** automation pipeline that syncs a Hardcover reading shelf into
Shelfmark requests and (optionally) downloads them, skipping anything already in an
Audiobookshelf library. Everything is **disabled by default** — with all toggles off,
behavior is identical to today.

## What this adds
A new **Settings → Hardcover Sync** tab with three independently-toggleable capabilities:

- **Hardcover wishlist sync** — pull a Hardcover shelf (e.g. "Want to Read") into
  Shelfmark as requests on a schedule (configurable in minutes or hours) or via a
  "Sync now" button. Synced requests carry full metadata and cover art.
- **Automatic downloads** — auto-approve and download synced requests, choosing the best
  release from a drag-to-reorder **source-priority** list. A strict title/author/
  audiobook-format (and seeder) match guard avoids grabbing the wrong file; anything
  without a confident match is left as a pending request for manual review.
- **Library-aware de-duplication** — check an Audiobookshelf library (ISBN + fuzzy
  title/author) and skip books already owned. Fails open (proceeds) if ABS is
  unreachable, so the pipeline never stalls.

## Design
- **Opt-in, no default behavior change** — all features gated behind settings that
  default to off.
- **Reuses existing systems** — synced books flow through the normal request →
  `fulfil_request` → `queue_release` download path; there's no parallel pipeline.
- **One behavior-preserving refactor** — extracts the per-source release search out of
  the `/api/releases` route into a reusable helper so the auto-picker and the route
  share it.
- **Conservative matching** — biased toward leaving items pending rather than
  downloading the wrong file.
- **In-process scheduler** modeled on the existing download coordinator (daemon thread).

## New modules
`core/text_match.py`, `core/release_search.py`, `core/auto_download.py`,
`core/hardcover_sync.py`, `core/hardcover_scheduler.py`, `core/library_index.py`,
`config/hardcover_sync_settings.py`; plus wiring in `main.py` (scheduler start + search
refactor) and `config/settings.py` (tab registration).

## Testing
- `ruff` clean on all changed files.
- Existing `tests/core` suite (request/release/search/requests-service) passes.
- Verified end-to-end against a live Hardcover account and Audiobookshelf instance:
  wishlist → request (with cover) → strict-matched auto-download, plus owned-book skip
  and fail-open when ABS is down.

## Scope note
[...]

PR #1355, opened 2026-09-19
refactor: extract the per-source release search out of /api/releases

splitsec2:
The `/api/releases` route carries an inner `_search_source_releases` helper that builds the search plan for one source, logs the planned query type, runs the search and turns `SourceUnavailableError`/operational errors into an error message instead of raising. Anything outside the route that wants to search one source with exactly those semantics has to go through Flask today.

This moves that helper into `shelfmark/core/release_search.py` as `search_source_releases()` and has the route delegate to it. Behaviour is unchanged: same plan construction (including the caller's `user_id`, so per-user default languages still apply), same logging, same error-to-message handling.

It is the refactor half of #1047 by @InfiniteAvenger, split out on its own as you asked for other PRs (#1318). Their authorship is preserved on the commit; I rebased it onto current `main` and added tests.

## Verification

- `tests/core/test_release_search.py`: unknown source → `"Unknown source: …"`, `SourceUnavailableError` and operational errors → `"<source>: <error>"`, success path forwards `expand_search` / `content_type` and returns the source instance, the plan receives languages / manual query / indexers / `user_id`. (These tests are type-annotated; happy to strip the annotations if you prefer the suite's bare style.)
- Full suite, ruff, ruff format, basedpyright, vulture green; the existing `/api/releases` route tests are unchanged and pass.


calibrain:
LGTM !

PR #1357, opened 2026-09-20
fix(download): check task ownership before serving queued files

splitsec2:
`/api/localdownload` resolves the file through the live queue and returns it before checking who owns the task; the owner check only runs on the download-history fallback, once the task has aged out of the queue. Task ids are source ids, so two users who searched the same book can end up with the same id.

This applies the same rule on the queue path, reusing the `_task_owned_by_actor` helper the cancel/retry/priority routes already use, so both paths answer a non-owner with the same 404. Admin behaviour is unchanged.

The 404 matches what this endpoint's history path already returns for a non-owner rather than the 403 `download_not_owned` the cancel/retry/priority routes use, happy to switch it if you prefer consistency with the siblings instead.

## Verification

- `tests/core/test_activity_routes_api.py`: the owner still receives their queued file; a different user receives 404. The new case fails on current main and passes here; the existing history-fallback test is unchanged.
- Full suite (3148), ruff, ruff format, basedpyright, vulture green.


calibrain:
Thank you !

splitsec2:
Adding the impact now that this is merged, so anyone reading it later can tell whether it affected them.

On a multi-user instance, any logged in user could download another user's finished file while the task was still in the live queue. That window is `STATUS_TIMEOUT`, one hour by default. Task ids are source ids, so a second user who had searched the same book already held the id and did not have to guess anything. Nothing else leaked, and a single user instance was never affected.

The fix is in main but not in a release, so v1.3.15 still has it. Worth a release when you have one queued.

splitsec2:
One process note and then I will stop adding to a merged PR.

There is no private way to report this kind of thing today. Private vulnerability reporting is off and there is no SECURITY.md, so I sent both of these as ordinary bug PRs and left the impact out until they were merged. If you switch on private reporting under Settings > Code security, the next person who finds something gets a channel that is not a public issue, and you can publish an advisory straight from the report once the fix ships. It costs nothing while nobody uses it.

Happy to send a short SECURITY.md as a PR if you want one.

PR #1358, opened 2026-09-20
fix(download): default is_admin to False in the request policy guard

splitsec2:
`_resolve_policy_mode_for_current_user` reads `session.get("is_admin", True)`, so a session carrying `user_id` but no `is_admin` key skips the request policy entirely, while every other admin check in the codebase defaults the key to `False`.

This uses the same default here. Every authenticated login path (builtin, CWA, proxy, OIDC) writes `is_admin` into the session, and `AUTH_METHOD=none` is already short-circuited one line earlier, so sessions from those flows behave exactly as before.

## Verification

- `tests/core/test_request_routes_api.py::TestDownloadPolicyGuards`: a session without `is_admin` now gets `policy_requires_request` and nothing is queued. Fails on current main, passes here.
- Full suite (3147), ruff, ruff format, basedpyright, vulture green.


calibrain:
Thank you !

PR #1359, opened 2026-09-20
fix(oidc): reject backslash paths in the return_to sanitizer

splitsec2:
The OIDC `return_to` sanitizer rejects values starting with `//` and then relies on `urlsplit` to catch anything carrying a netloc. A value such as `/\host` has no netloc, so it is stored in the session and used as the post-login redirect target and browsers resolve the backslash as a path separator, which lands the user outside the app after a successful login.

`_normalize_return_to` now also rejects values whose path contains a backslash. That matches the frontend sanitizer in `authRedirect.ts`, which parses with `URL` and already discards those forms, so the two ends agree again. The check covers the path only, so query and fragment backslashes still round-trip, and it also catches the script-root case where `/app/\host` strips to `/\host`.

## Verification

- New cases in `tests/core/test_oidc_routes.py` cover the rejected forms, including under a script root, and confirm `/`, `/settings` and `/search?q=x#frag` are unaffected. They fail on current main and pass here.
- Full suite (3155), ruff, ruff format, basedpyright, vulture green.


calibrain:
Looks good !

splitsec2:
Same as #1357, adding the impact now that it is merged.

A crafted login link could drop a user on another site after a successful OIDC login, because the backend accepted a backslash path that browsers resolve as a host. It needs the user to follow the link, and no session or token is exposed, so the practical risk is a phishing page that starts from a trusted domain. Only instances using OIDC were affected, and the frontend sanitizer already rejected these values, it was only the backend redirect that did not.

Also in main only, not in v1.3.15.

PR #1362, opened 2026-09-20
Feature: Show the AA search result stats

RoninTech:
When doing a direct search for a book, show the stats of the AA results.  For example if we search for "The Great Gatsby", AA reports it has 240 hits and shows the first page of  50.  Provide this stats info in the shelfmark webUI via ResultsSection.tsx and ReleaseModal.tsx.  This table shows what should be displayed based on the total number of hits found:

|Total	|Display|
|-------|-------|
|1	|"Result 1 (1 Total)"|
|6	|"Results 1-6 (6 Total)"|
|144	|"Results 1-50 (144 Total)"|
|500+	|"Results 1-50 (500+ Total)"|

Currently, shelfmark also only shows us the first 50 search hits even if there were more available from AA.  This could be added later if considered desirable.

As usual, a picture is worth a 1000 words:

direct-results-stat

universal-results-info


calibrain:
Looks great ! Thank you very much !

PR #1371, opened 2026-09-21
feat(search): add a configurable default content type

### LINKED ISSUE #1018: [Feature Request] Configurable default content type (ebook vs. audiobook)
**Problem**
The search page always opens with the ebook tab selected. This is hardcoded in SearchSection.tsx (contentType = 'ebook') with no setting to change it. Users who primarily use Shelfmark for audiobooks have to manually switch the tab on every visit.

**Proposed Solution**
Add a DEFAULT_CONTENT_TYPE setting that controls which tab is selected on page load.
This would fit naturally alongside the existing per-user search preference overrides already exposed in the Search Preferences panel, and could support both a global default and a per-user override like the other settings there.

**Current Workaround**
The URL parameter ?content_type=audiobook already works for this on a per-visit basis — it just needs to be surfaced as a persistent setting.

splitsec2:
Closes #1018.

Worth correcting the issue first: the search tab is not hardcoded. `useContentTypePreferences` has persisted the user's choice to localStorage since #564, so a browser that has picked a tab already keeps it. What is missing is the other half the issue asks for, a default for a browser that has stored nothing, and a per-user override.

`DEFAULT_CONTENT_TYPE` is a user overridable select next to `BOOK_LANGUAGE`, so it follows the same path: global value in Settings, per-user value in Search Preferences, resolved with `user_id` in `/api/config`. The frontend uses it only when this browser has no stored choice, which is captured before the existing effect writes one, so nothing changes for anyone who has already picked a tab.

The resolution is a pure function in `utils/contentTypePreference.ts` rather than logic inside the hook, since vitest here has no jsdom and the existing tests cover resolvers like `resolveDefaultLanguageCodes` the same way.

One small move in `App.tsx`: the `config` state was declared below the hook that now reads it, so it moved above it.

## Verification

- `tests/core/test_config_api.py`: the payload carries `default_content_type` and reads it with the user's id.
- `tests/core/test_admin_users_api.py`: the key appears in the per-user search preferences list.
- `src/frontend/src/tests/contentTypePreference.test.ts`: a stored tab wins, combined mode survives, the server default applies when nothing is stored, and an unrecognised value falls back to ebook.
- Python suite (3163) and frontend suite (201) green, plus ruff, ruff format, basedpyright, vulture, tsc, oxlint, oxfmt and the production build.


calibrain (changes requested):
Thank you very much, this is a great idea !
There are 2 issues before we can merge ;

Deep links break on a first visit. The new effect runs after the URL-search bootstrap in the same render and overwrites its content type. The link's search still runs, so audiobook results end up under the Ebook tab. This happens even when the setting is left at its default, and it breaks the ?content_type=audiobook workaround that issue [#1018](https://github.com/calibrain/shelfmark/issues/1018) mentions.

The default rarely takes effect. The existing persist effect saves 'ebook' to the browser on the very first page load, before the config arrives. On OIDC instances the sign-in round-trip reloads the page, so the stored 'ebook' then blocks the default permanently. Every browser that has opened Shelfmark since [#564](https://github.com/calibrain/shelfmark/issues/564) is also affected, so existing users will never see the setting.

splitsec2:
Thank you for catching both of these. They are real and they are my fault. I applied the configured default as a late effect and left the old eager persist in place, which is exactly what made both failures possible. I noticed that first-load write while I was building it and decided not to touch it, and that was the wrong call.

Pushed f5f993a.

The preference now records where it came from. Storing a tab, or anything explicitly setting one, marks it chosen, and only a chosen tab is persisted or allowed to outrank the configured default. So the URL bootstrap marking the tab chosen is enough for the default effect to leave it alone, and `?content_type=audiobook` lands on the Audiobook tab again. The default now also applies on every load until someone actually picks a tab, instead of being shut out by a value nobody chose.

For the browsers that already carry a value, the storage key moves to `preferred-content-type.v2`. Those old values were written automatically rather than picked, so honouring them would keep the setting shut out permanently. The cost is that anyone who had deliberately chosen a tab gets reset once, and their next pick sticks. If you would rather not reset anyone, the alternative is that the setting only ever reaches new browsers, and I am happy to switch to that instead. Your call.

Tests cover the decision table, including that a configured default stays overridable while a stored choice does not. Python and frontend suites, ruff, basedpyright, vulture, tsc, oxlint, oxfmt and the build are all green.

PR #1383, opened 2026-09-22
Feature: Add the ability to process more than 1 page of AA search results

RoninTech:
Now that we are showing how many AA results are actually being displayed it is front and centre that the default is to ignore AA results beyond 50.  ABB searches already have a config item to allow for processing more than 1 page of search results.  This PR adds a new config item for `Max Anna's Archive Results Pages` that defaults to 1 to match the original AA result handling, and allows for more than 50 results when it is set to > 1.  The maximum allowed pages to process is 10 which matches AA's search results, which only tell us there are 500 or more results.

The `Max Anna's Archive Results Pages` default of 1 will perform the same as current AA direct searches.  Increasing it will obviously increase the time for a search to return but can provide a larger set of the original AA paginated results.

Here are a couple of images to show the change:

Direct-dl-settings-max-AA-pages

direct-max-3-pages-AA-downloads


RoninTech:
There is a slight misalignment of the release number column, not perfectly centred between book cover and title/author, which is beyond my limited html/css abilities. Hopefully that's an easy fix @calibrain. 🙂

